PR #37, opened 2026-08-16
feat(auth): add opt-in TOTP 2FA, single active session, sign-in histo…

AbdallahMohamedDotnet:
…ry, and recovery-code alerts

Adds four independent, per-account, opt-in security toggles surfaced in a new Settings -> Security tab, each off by default:

- TOTP two-factor authentication (authenticator app + one-time recovery codes), enforced via a new two-step /auth/2fa/verify challenge after password or Google login succeeds.
- Single active session: a new login supersedes the account's previous session, independent of whether 2FA is enabled.
- Sign-in history: a bounded, newest-first record of past logins.
- Recovery-code alert: flags a login that used a recovery code instead of a normal TOTP code, surfaced on /auth/me and acknowledgeable from the Security tab. Only settable once 2FA is enabled, since recovery codes are a sub-concept of 2FA.

An account that leaves all four toggles off (the default for every existing account) sees byte-for-byte identical behavior to before this change: no server-side session-registry lookup, no history file, no alert field, and its 30-day stateless session cookie is unchanged.

Backend:
- Extract a generic signedPayload[T] HMAC sign/verify/expire primitive from SessionCodec (behavior-preserving), reused for real sessions, pending 2FA-login challenges, and pending enrollment tokens.
- Add stdlib-only RFC 6238 TOTP and recovery-code primitives.
- Add TwoFactorAuthenticator and SessionRegistry sub-components, composed into serviceauth.Service alongside the existing LocalAdminAuthenticator/GoogleAuthenticator sub-components.
- Add filetwofactor/filesessions file-backed stores following the existing hash-keyed-filename, temp+rename store shape.
- CurrentSession gains a ctx parameter and one more account-scoped rule (single-session supersession), applied for free to every existing caller (PrincipalResolver, AccessVerifier, skills catalog).
- Replace SignSession with IssueSession and CompletePasswordLogin / CompleteGoogleLogin / CompleteTwoFactorChallenge, which own the verify-then-decide-2FA-branch login policy so handlers stay thin.
- Add /auth/2fa/verify, /auth/2fa/cancel, and /api/me/security/* handlers for enrollment, disable, recovery-code regeneration, the three independent preference toggles, and alert acknowledgement.

Frontend:
- Add a 2FA challenge step to the login flow and a Security settings tab (enroll/confirm/disable, QR + manual-entry secret via the new qrcode dependency, recovery codes shown once, session history, preference toggles, alert banner).

Docs:
- Update ARCHITECTURE.md, docs/known-limitations.md, docs/threat-model.md, and docs/02-workspaces/02-auth-users-and-access.md to describe the opt-in tracked-session model.

Implements the plan in
docs/plans/2026-08-16-totp-2fa-single-session-device-history.md.


kingsoftheweb:
@AbdallahMohamedDotnet can you merge [futrx-com:qa](https://github.com/futrx-com/remote.futrx/tree/qa) into your branch, and fix the conflicts?

docs/plans/2026-08-16-totp-2fa-single-session-device-history.md
```diff
@@ -0,0 +1,229 @@
+# TOTP two-factor authentication, single active session, device history, and recovery-code alerting
```
kingsoftheweb: @AbdallahMohamedDotnet  we don't this

backend/internal/transport/http/handlers/auth_cookie.go
```diff
@@ -6,11 +6,43 @@ import (
 	serviceauth "github.com/futrx-com/remote.futrx.com/internal/service/auth"
 )
 
-func setSessionCookie(w http.ResponseWriter, auth *serviceauth.Service, user serviceauth.User) {
+// pendingTwoFactorCookieName carries the short-lived pending-login token
+// between a completed first factor (password/Google) and the 2FA challenge
+// endpoint. 5 minutes matches the pending-login token's own TTL.
+const (
+	pendingTwoFactorCookieName   = "remote_2fa_pending"
```
kingsoftheweb: @AbdallahMohamedDotnet  you are introducing a new cookie, it should also be added to the list of stripped cookies in infra/templates/Caddyfile.tmpl
for example: https://github.com/futrx-com/remote.futrx/blob/123bf572e40ba256a3ce2c7b87e6602a66b480d9/infra/templates/Caddyfile.tmpl#L69

Also it is better to keep the cookies constants here: remote.futrx/backend/internal/service/auth/model.go

kingsoftheweb (changes requested):
@AbdallahMohamedDotnet  one thing also needs to be done:

upon user deletion, we need to remove everything totp related to them

https://github.com/futrx-com/remote.futrx/blob/qa/backend/internal/service/user_removal_cleanup.go

which gets called here:
https://github.com/futrx-com/remote.futrx/blob/qa/backend/internal/service/services.go#L173

frontend/src/ui/primitives/QrCode.tsx
```diff
@@ -0,0 +1,38 @@
+import { useEffect, useRef } from "preact/hooks";
+import { toDataURL } from "qrcode";
```
kingsoftheweb: @AbdallahMohamedDotnet  you don't need a library that will add to our dependencies list, to generate qr codes.

You can add one JS class that can generate qr codes into a <canvas> element in the dom on the fly.

You can ask any ai agent to create a QRGenerator class.

here is an example : https://dev.to/dev_nestio_229945f10652e4/i-built-a-qr-code-generator-in-pure-vanilla-js-no-libraries-no-server-202-tests-7h

backend/internal/integration/agents/runtime/process.go
```diff
@@ -118,8 +118,10 @@ func RunProcess(
 		})
 	}
 
-	err = cmd.Wait()
+	// Drain stderr before Wait: Wait closes the pipes returned by StderrPipe,
+	// which can truncate an in-flight read in the goroutine above.
 	stderrText := <-stderrDone
+	err = cmd.Wait()
```
kingsoftheweb, line 124: @AbdallahMohamedDotnet  this was a bug you faced ?

backend/internal/service/auth/model.go
```diff
@@ -43,19 +47,119 @@ type Session struct {
...
+	IssuedAt  int64        `json:"issuedAt"`
+}
+
+// SessionHistory is the bounded (newest-first), bounded-size list of past
+// sign-ins for one account. Only populated while HistoryEnabled is on.
+type SessionHistory struct {
+	Entries []SessionRecord `json:"entries"`
+}
+
+const sessionHistoryCap = 20
```
kingsoftheweb: @AbdallahMohamedDotnet  move all the similar configs to 

https://github.com/futrx-com/remote.futrx/tree/qa/backend/internal/config

kingsoftheweb (changes requested):

backend/internal/service/auth/session_registry.go
```diff
@@ -0,0 +1,231 @@
...
+	record, err := r.load(ctx, email)
+	if err != nil {
+		return err
+	}
+	if record == nil {
+		return nil
+	}
+	updated := *record
+	updated.ActiveSessionID = ""
+	if err := r.store.Save(ctx, email, updated); err != nil {
```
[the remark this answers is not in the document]
kingsoftheweb: @AbdallahMohamedDotnet this is correct. all older session cookies wont have this id

backend/internal/transport/http/handlers/auth_twofactor_handler.go
```diff
@@ -0,0 +1,74 @@
...
+		httptransport.SendErr(w, http.StatusBadRequest, "invalid request")
+		return
+	}
+
+	key := localClientIP(r) + "|2fa"
+	if !h.limiter.Allow(key) {
+		w.Header().Set("Retry-After", "300")
+		httptransport.SendErr(w, http.StatusTooManyRequests, "too many attempts; try again in a few minutes")
+		return
+	}
```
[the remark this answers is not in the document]
kingsoftheweb: @AbdallahMohamedDotnet  you can fix this too

PR #41, opened 2026-08-17
feat(deploy): backup/restore tooling, small-host swap, root linger, launch timeout, repo URL override

mustafzidan74:
## Summary
Operations hardening for self-hosted installs, all opt-in-safe and idempotent:

- **`remote-backup` / `remote-restore`** (`infra/backup/`) + nightly `remote-backup.timer` (03:30 UTC ± 20 min) installed by new step `08-backup.sh`. A snapshot captures `DATA_DIR` (backend quiesced for the seconds it takes to tar — containers, previews and IDEs stay up thanks to `KillMode=process`), the project workspaces (live), provider tokens + Caddyfile + systemd units + LXD profile, and optionally `lxc export` of every container. Checksums + manifest, daily/weekly retention, optional `rclone` offsite copy via `/etc/remote-backup.env`. `remote-restore` verifies checksums, moves the current dirs aside as `*.pre-restore-<ts>`, re-chowns workspaces to the container idmap and restarts. Addresses the "No backup or restore mechanism" limitation.
- **`09-host-swap.sh`**: 4 GiB swapfile (swappiness 10) on hosts with < 8 GiB RAM and no swap, so one busy container degrades instead of OOM-killing the host.
- **`loginctl enable-linger root`** in `04-backend-svc.sh`. The snap-packaged `lxc` client attaches to root's *user* manager whenever an SSH session exists; when that session logs out, `user@0.service` stops and SIGTERMs every in-flight lxc process owned by the backend — observed as `lxc init: signal: terminated` mid image-unpack, leaving a half-created instance. Lingering keeps the manager alive.
- **Launch timeout**: `lxc init` ceiling raised from 90 s to 5 min (`FUTRX_LAUNCH_TIMEOUT` overrides). Unpacking the multi-GiB base image onto a dir pool takes >90 s on a 1-vCPU host and the timeout strands a broken container.
- **`FUTRX_REPO_URL`** lets `install.sh` clone/track a fork.
- **`.gitattributes`** enforcing LF: Go-embedded provisioning scripts built on Windows otherwise ship with CRLF.

## Testing
Deployed on a Hostinger 1 vCPU / 4 GiB Ubuntu 24.04 host: backup (1 s downtime) + restore round-trip verified, timer active, swap active, linger fixed the SIGTERM reproduction, `install.sh` re-run is idempotent. Docs: operations guide section + known-limitations update.


mustafzidan74:
This one's still clean against qa, no conflicts. Ready whenever you want to take it.

PR #42, opened 2026-08-17
feat(projects): public share links for app previews

mustafzidan74:
## Summary
Public preview links, so a project member can show a running app to someone without inviting them as a platform user (closes the "No public/anonymous sharing" limitation).

- **Project → Sharing → Public preview links**: per discovered port, pick 1h / 24h / 7d and **Share**. The URL `https://<slug>--<port>.dev.<host>/?share=<token>` is shown once with copy; live links list with remaining lifetime + **Revoke**.
- Backend: `internal/service/share` + `stores/fileprojectshares` (`DATA_DIR/projectshares/<id>.json`, stores only a SHA-256 of the token). Routes `POST/GET /api/projects/{id}/shares`, `DELETE …/shares/{shareId}` (membership required).
- Edge auth: `/auth/verify` on `*.dev.<host>` accepts a valid `?share=` token for that exact slug+port, answers 302 that sets an HMAC-signed `remote_share` cookie **scoped to that preview hostname only** (never the base domain) and strips the token from the URL. Cookie path is re-validated per request so revoke is immediate. Never grants the IDE (`*.code.`), the main app, another port/project, and platform plumbing ports (6080 noVNC, 8842/8081 code-server, 9222 CDP) are refused outright.
- Session codec gained a domain-separation purpose; session signing is byte-identical so existing cookies keep working.
- Caddyfile template: `remote_share` added to the cookie-strip list (defense in depth; needs a re-render on update).

## Testing
Unit tests for store/service/verify handler/frontend state. On a live host: anonymous link → 302 + host-scoped cookie → 200 from the app; wrong port/IDE/main app rejected; revoke → next request bounces to login; on-disk file has no plaintext token.


kingsoftheweb:
@mustafzidan74 
can you merge [futrx-com:qa](https://github.com/futrx-com/remote.futrx/tree/qa) into your branch, and fix the conflicts?

Ahmedwaleed22:
@mustafzidan74 
I suggest adding an agent skill to configure Cloudflare Tunnel: https://trycloudflare.com/
So we don't have to deal with any authentication like ?share=<token> since we may be sharing a full webapp and we cannot attach a token parameter in every single request

mustafzidan74:
You're right that a token in the query string falls apart for a full webapp, it only survives the first request.

That's why the token isn't the session. The first hit exchanges it for a signed cookie scoped to that one preview subdomain and port, and every request after that carries the cookie, so links and XHR inside the app work without the URL ever mentioning a token. auth_verify_share.go does the exchange, and the cookie is deliberately set with no Domain so the browser can't widen it past that host.

On the tunnel, I'd rather not make it the default. These are unreleased client sites and a tunnel means a third party terminating TLS on them. As an opt-in skill for people who want a link that works without any auth at all, sure, that's a different tradeoff and a reasonable one to offer.

mustafzidan74:
Merged qa in, conflicts fixed, mergeable again.

Three of them. services.go picked up the agent runtime rework so I took qa's struct and put the Shares field back on it. stores.go was just both imports. And package.json, where qa swapped the explicit test file list for node's own discovery, so I dropped the list rather than merging it.

PR #43, opened 2026-08-17
feat(notify): Telegram and webhook notifications for agent-run lifecycle

mustafzidan74:
## Summary
Outbound notifications so an operator who walks away from a run gets pinged.

- **Settings → Notifications** (admin): Telegram bot (token + chat id) and/or generic JSON webhook (URL + shared secret, `X-Remote-Signature: sha256=<hmac>` over the body), per-event toggles for *run finished / run failed / needs attention / scheduled task*, and a **Send test** button reporting per-sink results.
- `internal/service/notify`: bounded queue, single worker, 3 retries with backoff, 10 s HTTP timeout, one "finished" per run — never blocks or fails the run pipeline. Hooked via a `RunObserver` option on the prompt and schedule services; "needs attention" fires when the agent invokes a human-handoff tool (`AskUserQuestion`, `ExitPlanMode`, `RequestPermission`, …).
- Every message carries a deep link `<BASE_URL>/?chat=<id>`; the SPA consumes that query param once on boot (first URL-addressable thing in the app — small `chatDeepLink` state).
- Config in `DATA_DIR/notifications.json` (0600); token/secret are masked in GET responses and retained on save when omitted; clearing a destination also drops its retained secret.

## Testing
httptest-based notifier/queue/retry tests, store masking round-trip, handler auth. On a live host: real Claude run → webhook received with valid HMAC and working deep link. Docs: `docs/02-workspaces/07-notifications.md`.


kingsoftheweb:

mustafzidan74:
Merged qa in and fixed the conflicts, mergeable again.

The agent layer rework touched most of what conflicted, so services.go, transport.go and server.go are qa's versions with the notifier wiring put back on top rather than the other way round. Behaviour is unchanged. Two smaller ones worth mentioning: qa already ships a notifications settings tab so SettingsPage took its version outright, and we'd both added a Bell icon, so I kept qa's since it sits next to the BellOff it pairs with.

kingsoftheweb:
@mustafzidan74  like the idea of adding external notifications. But I want to make sure this could be well integrated with the push notifications service we introduced lately.

Unless this is supposed to satisfy a different need?

mustafzidan74:
Different need, and the difference is who it reaches.

Push is per-user: it goes to browsers that have visited, granted permission, and still have a live service worker, and the audience is the chat's members with presence filtering so the person already looking at the chat doesn't get buzzed. That's exactly right for what it does.

This is one config per deployment, pointed at the operator's own channels. It isn't trying to notify users — it's trying to reach me when I'm not at the machine.

The clearest way to see they're not the same thing is the events that have no chat and no browser to attach to:

- `siteWatch`: a client's live website went down. Nobody has the platform open at 3am, and there is no chat for it.
- `system`: the backend process came up. That fires precisely when there is no session to push to.
- `projectHealth`: a container crossed a memory or health threshold, which happens whether or not anyone is signed in.
- `digest`: the weekly cost roll-up, which is a thing you read, not a tray entry that gets replaced by the next one.

And the webhook sink isn't a human destination at all. It's how a run result gets into Slack, a status board, or someone's own automation. Push can't do that by design.

Where you're right that they overlap is `runFinished` and `needsAttention`. Both of us hang off the same moment, and `needsAttention` is literally the same trigger — your notifier watches for the `AskUserQuestion` tool and so does mine.

So I'd rather not merge the two services: their audiences resolve differently (per-user vs per-deployment) and one of them can't be delivered without a browser. What I think is worth sharing is the policy above them — one place that decides an event deserves a human's attention and at what urgency, with push and the sinks as two transports under it. Right now that judgement is written twice.

Happy to do that as a follow-up on top of this, or to restructure this PR that way first if you'd rather it land already shared. My preference is the follow-up, since this one is mergeable as it stands and the shared-policy change would touch your push code — better as its own reviewable diff. Say which and I'll push it.

PR #44, opened 2026-08-17
feat(audit): append-only audit log with admin UI and export

mustafzidan74:
## Summary
Append-only audit log — who did what on the box (closes the "No audit logging" limitation, and removes the `callerEmail // available for future audit logging` placeholder).

- `internal/service/audit` + `stores/fileaudit`: monthly JSONL files at `DATA_DIR/audit/audit-YYYY-MM.jsonl` (0600, `O_APPEND`), one object per line `{at, actor{email,sub,isAdmin}, action, target{type,id,name}, ip, userAgent, meta, ok, error}`. Writes never fail the caller. Read API filters by actor / action prefix / target / time range with cursor paging, reading only the needed monthly files. Daily janitor applies `AUDIT_RETENTION_MONTHS` (default 12, 0 disables).
- Instrumented (service layer via `WithAudit` options; actor propagated through request context by the auth middleware): sign-in success/failure, logout, admin claim, user invite/remove/role, project create/rename/delete, membership, secret read/set/delete, container start/stop/restart/recycle/repair/limits, agent-browser start/stop, chat create/delete, agent run start/cancel, schedules, Google-OAuth config, provider connect/cancel, self-update, workspace file upload/download/archive, git checkout, IDE/terminal opens.
- HTTP: `GET /api/admin/audit?actor&action&target&from&to&limit&cursor`, `GET /api/admin/audit/export` (NDJSON download). **Settings → Audit log** (admin): table, filters, load-more, export.

## Testing
Store append/read/filter/pagination/rotation/retention, middleware actor propagation, handler gating, and an integration test proving project create emits an audit line. Verified live (login failure, project lifecycle recorded). Docs: `docs/04-operations/10-audit-log.md`, env table, known-limitations.

PR #45, opened 2026-08-17
feat(projects): stack templates (blank, wordpress, laravel, node, python)

mustafzidan74:
## Summary
Project templates: create a container from a stack preset instead of only the bare base image.

- A template = the shared `futrx-remote-dev-base` image + a marker-gated post-provision script run once inside the new container + optional `/workspace` seed files + an `AGENTS.md` snippet (never overwriting). Optionally a template declares a prebuilt alias `futrx-remote-<name>-base`; if present it is used (fast path), else base + script (slow path). New `cmd/build-template-image -template <name>` builds the alias reusing the image builder (with `-publish-timeout`/`-build-timeout` overrides for slow hosts).
- Ships `blank` (default, unchanged behaviour), `wordpress` (PHP 8.3 + MariaDB + WP-CLI + Composer, WordPress core in `/workspace/public`, preview on :8080), `laravel`, `node`, `python`. Provisioning is non-interactive, idempotent (`/var/lib/remote-template.done`), logs to `/var/log/remote-template.log`, runs in the background, and re-runs automatically after a container is replaced (`upgrade-workspaces` re-clones the rootfs) since the marker lives in the disposable rootfs. `infra/upgrade-workspaces.sh` gains `--rebake-templates`.
- Project meta gains `template` (omitted for `blank` → backward compatible). `GET /api/templates`; project create accepts `template`. New-project dialog replaces the old `prompt()` with a template picker; project page shows template + provisioning status badges and a Template panel.

## Testing
Unit tests for catalog/validation, provisioning decision logic, meta compat, handlers, lifecycle wiring. Verified live on a 1-vCPU host: WordPress template provisioned in ~70 s, MariaDB + WP-CLI active, preview serves the WordPress installer; Node template idempotent across restarts. Docs: `docs/02-workspaces/08-project-templates.md`.

### DIFF
--- a/backend/cmd/build-template-image/main.go
+++ b/backend/cmd/build-template-image/main.go
@@ -1,0 +1,174 @@
+// build-template-image publishes a dedicated LXD image for one project
+// template — the optional FAST PATH described in
+// docs/02-workspaces/08-project-templates.md.
+//
+// Without a template image, a project created from e.g. the WordPress template
+// launches from futrx-remote-dev-base and installs PHP, MariaDB, and WP-CLI
+// inside its own container on first start (minutes on a small host). With one,
+// the lifecycle launches straight from futrx-remote-<template>-base and the
+// provisioning is already baked in.
+//
+// Usage:
+//
+//	go run ./cmd/build-template-image -template wordpress
+//	go run ./cmd/build-template-image -template wordpress -overwrite
+//	go run ./cmd/build-template-image -list
+//
+// Publishing compresses the whole rootfs and regularly exceeds the default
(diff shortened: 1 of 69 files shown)


kingsoftheweb:

mustafzidan74:
Same situation as #44, so flagging rather than pushing a bad merge.

qa has its own create-project flow now, CreateProjectModal plus createProjectForm. This branch has a new-project dialog with a template picker in it. They're two implementations of the same screen, so the merge is a rewrite: your modal is the one to keep and the template picker needs rebuilding inside it, with createProject taking a template alongside the name.

That's fine by me, just wanted to check you'd want the picker in the modal at all before I redo it there.

PR #46, opened 2026-08-17
feat(skills): platform-wide global skills library

mustafzidan74:
## Summary
An admin-managed **global skills library** injected into every project.

- `DATA_DIR/skills-global/<name>/` in the same SKILL.md-directory format as project skills, with `_index.json` for policy flags (`alwaysOn`). Admin CRUD: `GET/POST/PUT/DELETE /api/admin/skills-global[/{name}]` (JSON files map or zip upload) + `POST …/import` from a project.
- Every project chat sees global skills merged into its picker (badged `global`, read-only). They are published into each container at `/workspace/.agents/skills-global/` and symlinked into `/workspace/.agents/skills/<name>` only when that slot is free — a project-local skill of the same name wins and the global one is reported `shadowed`. Sync runs at the top of `EnsureSkillLinks` (container launch + every agent run) guarded by a library hash, so a running container picks up changes on its next turn; removed skills are pruned.
- `alwaysOn` skills are preselected in every new project chat (default, not enforced — documented).
- Two built-ins embedded in the binary and seeded on first run only when the library is empty: `code-review-guard` and `wordpress-guard`.
- **Settings → Global skills** admin page; skill picker shows the badge.

## Testing
Store CRUD, collision/shadowing, listing merge, admin gating, seed idempotency, frontend state tests. Verified live: seeds installed, symlinks appear inside the container after a run. Docs: `docs/02-workspaces/09-global-skills.md`.


kingsoftheweb:

mustafzidan74:
Merged qa in, mergeable again.

Four of the conflicts weren't just picking a side. servicechat has its own Option type now so mine was a duplicate, and the field it sets moved onto your struct. skills.projectRoots became a method with a provider argument, but the global library reads a project's skills without a Service, so I put the provider-independent search order back as a package function and had the method call it. And two constructors grew arguments, which Go won't let you mix with a spread, so your options get appended to this branch's slice.

SettingsPage carries both tab sets and both icon imports now, since either side alone drops a real tab.

kingsoftheweb:
@mustafzidan74  how is this different to the global skills that get added to /workspace/.agents/  by default

Like the "browser" skill ?

mustafzidan74:
Different owner, different lifecycle.

`browser` and `scheduled-tasks` are yours. They're `//go:embed`ed in the binary and pushed into `/workspace/.agents/skills` at launch, re-pushed when the embedded file changes. They document platform capabilities — the browser one is the playbook for the `browser_*` tools. Nobody can add to that set without a release, and nobody should: they're part of the feature they describe.

This is the operator's set. An admin writes a skill once in Settings, it lands in `DATA_DIR/skills-global`, and every project gets it. Editing it takes effect on the next chat start, no deploy. On my install that's things like a WordPress review checklist and our own code-review guard — house rules that have nothing to do with the platform and that I don't want to paste into twenty projects.

They deliberately share a format and a directory. Same SKILL.md layout, and the library is mirrored to `.agents/skills-global` and linked into `.agents/skills`, so the agent CLI sees one namespace and a skill can be moved between a project and the library without translation. Precedence is project-local > global: the link is only created when the slot is free, so a project that wants its own version just adds one.

Your question made me check what happens when the two sets collide, and there was a real bug. Launch links the library in *before* pushing the built-ins, so a library entry named `browser` gets the built-in's SKILL.md written straight through the symlink and into the mirrored copy. I reproduced it in a container: `install -d` succeeds on the symlink and the write replaces the operator's file. The library's hash marker is untouched, so nothing repairs it until the library changes for some other reason.

Pushed a fix — both names are refused at authoring time now, with their own error so the operator is told the name is taken rather than malformed. If you add more provisioned skills later, that list is the place to keep in step.

PR #47, opened 2026-08-17
feat(usage): token and cost dashboard with per-provider normalization

mustafzidan74:
## Summary
Usage & cost visibility per project / chat / user / provider / model.

- Provider adapters normalize their native usage payloads (Claude `usage` + `total_cost_usd`; Codex token counts; Kimi/Antigravity model-only) into one shape carried on the `complete` chat event, so the ledger can be rebuilt offline.
- `internal/service/usage` appends a `UsageRecord` per completed run to `DATA_DIR/usage/usage-YYYY-MM.jsonl` (0600). Cost: provider-reported when available, else estimated from an admin-editable `DATA_DIR/usage/prices.json` (marked `estimated`); providers with no token data are recorded as **unknown, never zero**.
- API: `GET /api/usage/summary?from&to&groupBy=project|user|provider|model|day`, `GET /api/usage/records`, `GET /api/projects/{id}/usage`, admin `GET/PUT /api/admin/usage/prices`, `POST /api/admin/usage/rebuild` (re-scan chat logs; also `cmd/usage-rebuild`). Members see only their projects.
- **Settings → Usage**: range picker, KPI tiles, inline-SVG daily chart (tokens/cost), grouped table with per-chat drill-down, rebuild button; project page header shows a month-to-date line.

## Testing
Parser fixtures per provider, aggregation math, monthly rotation, membership filtering, price estimation, frontend state tests. Verified live with a real Claude run (exact `total_cost_usd` recorded). Docs: `docs/02-workspaces/10-usage-and-cost.md`.


kingsoftheweb:

mustafzidan74:
Merged qa in, conflicts fixed.

Three of them needed actual changes rather than picking a side. prompt.New grew two arguments, and since Go won't mix explicit variadic args with a spread I append qa's options onto this branch's slice. emitAgentEvent grew a provider argument. And agent.Provider gained Capabilities, so the usage test's provider double implements it and returns an empty set, which seemed more honest than inventing one for a double that only exists to record events.

Also dropped the AgentAuth field on Services since qa removed agentauth entirely, and SettingsPage now carries both tab sets, yours and this branch's.

kingsoftheweb:
@mustafzidan74   This is a great feature and a very good PR. I only committed one update for a miscalculation that happens with Codex. And similar APIs, it just needed some normalization to prevent re-calculating cached tokens.

 I would like us, though, to keep improving on this because right now we are running calculations and usage based on API, not the subscription, while most of the users—if not all of them—will be using subscriptions.
 
  So maybe we need to investigate how can we integrate this. So we can actually get the correct limits.  I'm not sure if that, but a good investigation here to be able to fetch the subscription limits will be very helpful. But this could be another PR.

 Also, for the pricing right now, it's hard-coded. I think we should be pulling the prices dynamically as they can change at any time.

mustafzidan74:
Thanks for the merge, and for the cached-token normalization. That double-count was mine.

On subscription limits: I went looking for this after your comment and the CLIs already tell us, we were just throwing it away. Neither exposes a command to ask, they volunteer it mid-run.

claude emits a top-level line:

```json
{"type":"rate_limit_event","rate_limit_info":{"status":"allowed","resetsAt":1787563200,"rateLimitType":"five_hour","overageStatus":"rejected","isUsingOverage":false}}
```

I have seen `five_hour` live on a Max subscription. `seven_day`, `weekly` and `utilization` are all in the pinned binary too, so the weekly window presumably arrives the same way, but I have not caught one yet to confirm the shape.

codex hangs it off `token_count` instead, as `rate_limits` with `primary_window` and `secondary_window`, each carrying `used_percent`, `window_minutes` and `resets_in_seconds`. Those field names are from the pinned binary rather than a captured event, since I do not have a Codex subscription to log in with. Worth someone confirming before building on it.

Two things this shape forces:

Every reading is a snapshot from whenever the last run happened, not a live figure, so anything showing it has to say how old it is. An idle install would otherwise present yesterday's number as current.

And claude sends a status with no percentage. Storing that as 0% would tell the operator none of their plan is used, which the CLI never said, so the absent case has to stay absent rather than become a number.

I have this running on my own install already, capturing the window per agent and showing it on the dashboard. Happy to open it as its own PR if you want it, it does not touch the ledger.

Agreed on hard-coded prices, same problem as the model ids going stale underneath us. Separate PR though.

kingsoftheweb:
Sounds good. I will wait for the pr!

PR #48, opened 2026-08-17
feat(resources): host-aware fleet defaults, per-project envelope UI, aggregate memory guard

mustafzidan74:
## Summary
Per-container resource policy becomes runtime operator configuration instead of compiled constants, plus a host-level admission guard.

- `DATA_DIR/resources.json` {defaults, hostReserve, maxProjectOverride, maxRunningContainers}; **derived from real host capacity on first run** (a 1 vCPU / 4 GiB host no longer gets a 4 GiB / 6 CPU default) and converged onto the managed LXD profile at startup and on change. Root-disk quota applied per container with storage-pool driver detection (`dir` pools report "unsupported" instead of failing launches).
- `GET/PUT /api/projects/{id}/resources` (members read, admins write; validated against the ceiling and host capacity; reports `appliedNow`/`needsRestart`; includes a live usage snapshot). Replaces the previous `PUT …/limits` route (single write path).
- **Aggregate guard**: before any launch/start, sum of memory ceilings of running containers + this one vs host memory − reserve; refuse with 409 (`?force=1` for admins, surfaced as "Start anyway"). Prevents the reboot-autostart OOM scenario. Admin `maxRunningContainers`.
- `GET/PUT /api/admin/resources`; **Settings → Resources** (host capacity, commitment bar, pool quota indicator, fleet defaults) and a richer per-project Resources panel with usage bars and inheritance labels.

## Testing
Table-driven derivation, validation, aggregate math, pool-capability parsing, handler auth. Verified live: profile converged, override applied hot to a running container, second container refused with 409 until defaults were lowered. Docs: `docs/02-workspaces/11-resource-limits.md` + limitation updates.


kingsoftheweb:

mustafzidan74:
Merged qa in, mergeable again.

Three worth calling out. qa builds the server-info service inline in the Dependencies literal now, so the variable this branch held for it is gone and HostCollector was already being passed anyway. The definitions/profiles pair that sat above the resource-policy block is gone too, since qa removed it. And two constructors grew arguments, which Go won't let you mix with a spread, so your options get appended to this branch's slice.

SettingsPage keeps both tab sets and both icon imports.

PR #50, opened 2026-08-18
feat(applications): installable images — a catalog of apps that install into a project, extend the UI, and run server-side code

Ahmedwaleed22:
Adds **installable images**: a catalog of applications a user can install globally or
into a single project. An image is a directory, not a code change — adding one requires
no edit to Remote. `NewRegistry()` walks the catalog at startup, validates every entry,
and the app appears in the Applications tab.

## The shape of an image

```
images/<id>/
  image.json     metadata: type, version, install fields, host tools it needs
  install.sh     provisioner — runs inside a container (a service on a port)
  plugin/        Go source — compiled and run on the host as a process
  ui/            browser extension — buttons, panels, popups
```

Any one of the three, or all three. `ui/` is what an image can add to the interface;
`plugin/` is what it can add to the server. Together they are the whole shape of a
feature that could otherwise only be added by editing this repository.

## What lands

- **Catalog + registry** — embedded via `//go:embed`, validated at load; service,
  store, installer, and host-port allocator behind it.
- **Applications UI** — global and per-project tabs, connection details and
  credentials on app cards, uniform user/password/host/port for every server.
- **UI extensions** — a **closed set** of named slots (`frontend/src/config/extensions.ts`)
  with declared context. An extension naming an unknown slot loses that one
  contribution and keeps running, so an image built against a newer Remote degrades
  on an older one instead of failing to load.
- **Backend plugins** — `plugin/` is compiled and run as a separate process; requests
  are forwarded to it. `pkg/appplugin` is the public contract and depends on nothing
  but the standard library.
- **Uploaded packages** — an admin can upload a `.zip` from Settings → Applications.
  It lands in `$DATA_DIR/app-packages/`, never in the checkout or the binary, so it
  survives an update. Downstream of the registry nothing can tell it from a built-in.
- **Host tools** — an image declares a host executable; `hosttools` fetches exactly
  that over HTTPS and refuses anything whose SHA-256 does not match. Tools install
  under the data directory, never `/usr` and never via the host package manager, so
  an operator who installs no applications keeps precisely the software they
  provisioned. This is why `restic` is *not* in `infra/steps/01-host-deps.sh`.
- **Versions** — `image.json` carries a required `version`. An instance records what
  it was installed from; when the catalog's version *differs* (in either direction)
  the install script runs again, so a deliberate rollback is a real rollback.
- **Extension events** — `remote.events.on("upload.completed", …)` with a synchronous
  `claim(work)`, letting an extension relocate a finished chat attachment before the
  composer sends. `s3disk` uses it to move attachments onto its mounted bucket.

## Shipping in the catalog

`postgresql` (16), `mysql` (8.0), `redis` (7), and `s3disk` (0.3.0) — the last one
combining all three halves: a bundled standalone Go FUSE mount in `container/`, a
host-side `plugin/` reaching the container over `lxc exec`, and a `ui/` contributing
Mount controls. The bundled playground/sandbox fixtures were removed in this branch.

## Scope and enforcement

Installing is the gate, not shipping: an extension loads only if the image is installed
globally or in a project the caller belongs to, **and** that instance is `running`.
Stop is a real off switch. The browser never decides this — `GET /api/applications/ui`
returns the list it is allowed to load. Global app management is admin-only; project
membership is re-checked per request; secret env values are redacted outside the
credentials route; a plugin cannot act as its caller (`Cookie`/`Authorization` are
withheld, the caller is supplied separately).

**Extension code runs on the main origin with the SPA's privileges — there is no
sandbox and none is implied.** The trust boundary is the build. Review a new or edited
`ui/` exactly as you review any other frontend change; reviewing `install.sh` carefully
while skimming `ui/` gets the risk backwards, because the script runs in a disposable
container and the extension runs in the user's session. Uploading a package is an act
of trust identical to merging a directory into this repo.

## Also in this branch

- Workspace idleness is now answered from inside the guest (`WorkspaceBusy`): shells
  with a workspace cwd, open workspace files, agents, and listening applications count
  as pins even at zero CPU; fails closed on a FUSE-backed `/workspace`. An app declares
  its own long-running daemon in `/etc/remote/workspace-idle.d` so it does not pin every
  workspace it is installed in.
- A visible project chat posts to `/api/chats/{id}/activity` every 30s and on visibility
  change, so the server can tell an open workspace from an idle one.
- fix(chat): an upload target lookup now fails instead of silently falling back to the
  chat cwd.
- fix(applications): redis failing to start from a malformed config include.
- CI builds, vets and race-tests the s3disk nested Go module, which `./...` skips silently.

## Docs

Seventeen documents under `installable-images/docs/` (the root symlink), covering
`image.json`, image types, install scripts, slots, the extension API, a build-a-plugin
tutorial, scoping, the HTTP API, the security model, backend plugins, uploaded packages,
and versions/upgrades.

## Known gaps

- `docs/10-fixtures.md` still documents the three removed playground/sandbox fixtures,
  and several other docs reference them. No code does.
- `docs/01-overview.md` and the opening of `docs/13-security-model.md` still state there
  is "no upload endpoint"; `docs/16-uploaded-packages.md` documents one. Doc 13
  reconciles this further down, doc 01 does not.

## Notes for review

Base is `main`, which is 6 commits behind `qa`, so this PR's diff also carries the
public share-links work (`feat(projects): public share links for app previews` and its
follow-ups). Merging `main` into `qa` first would shrink the diff to this feature alone.

## Verification

`go build ./...`, `go vet ./...`, and `go test ./...` clean; `npm run build` succeeds and
all 313 frontend tests pass, on the merge commit.


kingsoftheweb:
@Ahmedwaleed22  can you merge [futrx-com:qa](https://github.com/futrx-com/remote.futrx/tree/qa) into your branch, and fix the conflicts?

Ahmedwaleed22:
@kingsoftheweb done

kingsoftheweb:
@Ahmedwaleed22  please fix conflicts

Ahmedwaleed22:
@kingsoftheweb Done!

kingsoftheweb:
@Ahmedwaleed22  should this be closed now?

PR #59, opened 2026-08-23
feat(infra): provision swap sized from RAM and free disk

### LINKED ISSUE #58: Sudden ram spikes when using nx with a platform that has multiple running apps
## Summary

Running several apps at once in an Nx workspace produces a large, short-lived
memory spike. Total RSS goes from about 1.3 GB to about 7 GB in a few seconds.
When the machine survives it, memory returns to baseline within roughly a
second.

This machine has no swap configured, so the spike is not survivable here. It
triggers the OOM killer and takes down the dev servers.

The steady state fits in memory with room to spare. Only the peak is the
problem, and it is roughly 5x baseline.

## Environment

- Nx version: 23.1.1
- Node version: 22.23.2
- Package manager: pnpm
- OS: Ubuntu 24.04.4
- Machine: 2 cores / 4GB RAM, no swap

Ahmedwaleed22:
Running several dev servers at once produces a large, short-lived RSS spike (~5x steady state). On a swapless box that transient peak trips the OOM killer and takes the dev servers down. Add a swap file during host setup, sized from installed RAM, free disk, and total disk, so the kernel can absorb the spike instead of killing processes.

Sizing (lib/swap-provision.sh): <=2 GiB RAM -> 2x, <=8 GiB -> 1x, larger capped at 8 GiB; then fitted to what the disk can spare after a reserve of max(2 GiB, 20% of disk), rounded down to 512 MiB. Skips entirely when meaningful swap already exists or the disk has no room. Also writes an fstab entry and a vm.swappiness=10 drop-in so steady state stays in RAM.

Wired into steps/01-host-deps.sh before the memory-heavy Go and base-image builds. Pure sizing functions are covered by a deterministic test with mocked system calls.

fixes #58


kingsoftheweb:
@Ahmedwaleed22  can you merge [futrx-com:qa](https://github.com/futrx-com/remote.futrx/tree/qa) into your branch, and fix the conflicts?

Ahmedwaleed22:
@kingsoftheweb done

kingsoftheweb:
@Ahmedwaleed22  I like the approach. but this will only be set for new instances.

What about already installed ones? the only path is the update route

Ahmedwaleed22:
@kingsoftheweb After checking the updating process, I can see that steps/01-host-deps.sh is going to run, in which it will execute ensure_swap, which will do the job of creating the swap memory

**No changes needed**

PR #71, opened 2026-08-26
Import project from public repo 

itsamostafa:
Lets you seed a new project from a public git repo instead of always
starting empty. On first provisioning it clones the url straight into
the workspace directory on the host, the same way Git History already
shells out to git, then re-chowns it for the container.

Kept auth out of scope for now only plain https:// urls, no
credentials. A private or bad url fails fast with a clean message
instead of hanging on a prompt or leaking raw git output back to the
client. Private repos still work fine today via the Terminal with your
own git credentials.

Re-running provisioning against an already-seeded workspace (container
recreate, upgrade) just no-ops instead of re-cloning.

### DIFF
--- a/backend/internal/config/containers.go
+++ b/backend/internal/config/containers.go
@@ -18,6 +18,7 @@ import (
 	containerruntimeassets "github.com/futrx-com/remote.futrx.com/internal/integration/containers/runtimeassets"
 	containerscheduletools "github.com/futrx-com/remote.futrx.com/internal/integration/containers/scheduletools"
 	containerworkspace "github.com/futrx-com/remote.futrx.com/internal/integration/containers/workspace"
+	"github.com/futrx-com/remote.futrx.com/internal/integration/gitcli"
 	"github.com/futrx-com/remote.futrx.com/internal/integration/hostfs"
 	servicebrowser "github.com/futrx-com/remote.futrx.com/internal/service/container/browser"
 	servicecli "github.com/futrx-com/remote.futrx.com/internal/service/container/cli"
@@ -135,6 +136,7 @@ func NewContainerStack(
 		resources,
 		launchProvisioner,
 		profiles,
+		gitcli.NewCloner(),
 	)
 	inspectionAdapter := containerinspection.NewAdapter(
 		runner,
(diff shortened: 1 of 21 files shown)


ActArtech:
Hi @itsamostafa — nice, clean scope on this one (https-only, fail-fast, no-op on re-provisioning). I picked it up and got it merge-ready on top of current `qa`:

- **Rebased your 3 commits onto current qa** (~667 files of drift since your base). The conflicts came from the frontend refactor that landed since: your git-import UI now sits on the refactored `CreateProjectModal` (adapted to the current theme tokens), and your `createProject(name, gitUrl?)` signature is wired through the refactored `WorkspaceContext`. Frontend: tsc+vite build clean, 246/246 tests pass.
- **Backend**: the gitcli cloner and project-service changes pass `go build`/`go vet`/`go test` — 77/77 packages, `gofmt` clean. Nicely done on `GIT_TERMINAL_PROMPT=0`, the `--` separator before the URL, and keeping raw git output out of the client error.
- **One extra commit**: qa itself currently has a stray token in `backend/internal/service/auth/service_test.go:85` (introduced by `13e3772a` on Sep 4) that breaks `go vet`/`go test` compilation — CI doesn't run `go test`, so it's invisible there. I dropped that line on the branch so the suite runs; upstream will want the same one-line fix at qa tip regardless of this PR.

Branch (4 commits: your 3 signed, authorship preserved + the qa fix):
https://github.com/ActArtech/remote.futrx/tree/feat/import-project-from-git-url

Happy to hand it over however you and the maintainers prefer: you're welcome to push this branch to your PR, or I can open a replacement PR citing #71. Thanks for the solid work!

itsamostafa:
Hi @ActArtech 
Thanks for the rebase and especially for catching the service_test.go:85 breakage, I confirmed it's on qa tip from 13e3772a, and that go vet/go test stop there.

I've rebased #71 onto qa tip (9a0fc44f) myself and force-pushed, I wanted to work through the frontend refactor rather than take it as a patch. Same numbers you got: frontend tsc+vite clean, 246/246 tests, backend go build/vet clean, 0 test failures, gofmt clean.

I kept the qa fix out of this PR and put it on its own branch off qa, since it's unrelated to the feature. Will check with the maintainers on landing it separately.

PR #73, opened 2026-08-26
fix(installer): handle non-main default branches

maher-elattar:
I found this while installing Remote on my VPS, which runs as an LXC container under Proxmox. I used the documented command:

```bash
curl -fsSL https://remote.futrx.com/get | sudo bash -s -- yasmeenremote.mosalam.com
```

The installer found the existing checkout in `/opt/remote.futrx`, printed that it was pulling the latest version, and then failed because `origin/main` did not exist.

The problem happens because the repository default branch currently points to `qa`. A shallow clone follows that default and is configured to fetch only `qa`. Later, the installer runs a normal fetch and tries to reset to `origin/main`, but that fetch never creates the missing remote-tracking branch, so Git reports `origin/main` as an ambiguous revision.

This fix makes the production branch explicit in both paths: fresh installs clone `main`, and existing narrow checkouts fetch `main` directly before resetting to it. I also added regression checks so changing the GitHub default branch cannot silently break the public installer again.

I reproduced the issue on the affected VPS with a `qa`-only fetch refspec and confirmed that the fixed fetch creates `origin/main` at the exact upstream commit. The installer shell checks pass as well.

@kingsoftheweb Mazen, thank you for the great work on this project. Remote is genuinely useful, and I appreciate how much thought has gone into making the self-hosted setup approachable.

### DIFF
--- a/infra/install.sh
+++ b/infra/install.sh
@@ -90,7 +90,12 @@ if [ -z "$INFRA_DIR_PROBE" ] || [ ! -d "${INFRA_DIR_PROBE}/steps" ]; then
             else
                 # Existing shallow clones may follow a different default branch,
                 # so fetch main explicitly instead of relying on their refspec.
-                git -C "$LEGACY_TARGET" fetch --quiet --tags origin "$MAIN_REFSPEC"
+                # Persist that refspec too — otherwise remote.origin.fetch stays
+                # on the stale branch and a later plain `git fetch origin` (in
+                # update.sh, steps/00-checkout.sh, ...) silently stops refreshing
+                # origin/main.
+                git -C "$LEGACY_TARGET" config remote.origin.fetch "$MAIN_REFSPEC"
+                git -C "$LEGACY_TARGET" fetch --quiet --tags origin
                 git -C "$LEGACY_TARGET" reset --hard origin/main
             fi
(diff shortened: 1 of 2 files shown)


kingsoftheweb:
I agree. Useful fix and confirmed the immediate issue is real. One concern: the explicit main refspec fetch creates origin/main but doesn’t update remote.origin.fetch, which remains qa on affected clones. Later normal fetches in update.sh and steps/00-checkout.sh may therefore leave origin/main stale. Please persist the main fetch refspec for repaired production checkout

maher-elattar:
Good catch. You're right — the explicit `origin/main` fetch created the ref but left `remote.origin.fetch` pointed at whatever branch the original narrow clone followed, so `update.sh` and `steps/00-checkout.sh` would keep fetching that stale refspec forever and never touch `origin/main` again.

Fixed in b805cf4..9972f6b: both repair paths (legacy checkout and primary checkout) now run `git config remote.origin.fetch "$MAIN_REFSPEC"` before the fetch, so the refspec is corrected permanently, not just for that one pull. Added a static assertion in qa-scripts-test.sh so both call sites stay covered, and verified the fix end-to-end against a local bare repo: simulate a `--single-branch qa` clone, apply the fix, then confirm a later *plain* `git fetch origin` (i.e., what update.sh does) picks up new commits on main.

PR #74, opened 2026-08-26
fix(installer): use native LXD inside Debian LXC

maher-elattar:
After fixing the Git bootstrap problem, I hit a second installer failure on the same VPS. The server runs Debian 13 as an unprivileged LXC container under Proxmox.

The installer always installs LXD through Snap. That works on a normal VM, but not in this kind of container: Snap first failed because it could not create a SquashFS loop mount. After installing the FUSE fallback, it still failed because the Proxmox guest cannot see the host AppArmor security filesystem.

Debian already ships native `lxd` and `lxd-client` packages, and those packages work here without Snap's packaging layer. This PR detects an unprivileged LXC host and uses the native Debian packages in that specific environment. Normal Ubuntu/Debian servers keep the existing Snap path.

There is one important safety detail. Remote expects every project workspace to be an unprivileged LXD container, and its bind-mount ownership code relies on the fixed `1000000:65536` UID/GID allocation. The installer now verifies that the outer Proxmox container delegates that range before continuing. It does not silently set `security.privileged=true`, because doing that would weaken the main isolation boundary and break the existing ownership assumptions.

I found the two host limitations directly on the VPS: native LXD installed and initialized successfully, while the first workspace image launch exposed the missing subordinate ID range. The new tests cover LXC privilege detection, range validation, and the native-package installer contract.

@kingsoftheweb Mazen, I split this from the Git branch fix so the two issues can be reviewed independently.


kingsoftheweb:
@maher-elattar  i changed the base to qa. There is now a conflict as qa had a change for the install's first step here:
https://github.com/futrx-com/remote.futrx/pull/59

please sync qa into you and resolve the conflict

maher-elattar:
Synced qa in and resolved the conflict in `01-host-deps.sh` (3d356e8). Both changes are kept: the nested-unprivileged-LXC / native Debian package detection from this branch, plus your `snap list lxd` check from #59 in place of `command -v lxc` for the non-nested path, since the Ubuntu 24.04 `lxd-installer` shim defeats a `command -v lxc` check. Ran the full infra test suite after the merge — lxd-host-test and the rest pass.

PR #77, opened 2026-08-27
fix(image): retry the IPv4 egress pre-flight instead of probing once (#39)

### LINKED ISSUE #39: build-base-image: IPv4 egress pre-flight races the builder's DHCP lease on slow hosts (false 'Docker' error)
## Summary

On a slow / single-vCPU host, `cmd/build-base-image` fails deterministically at the IPv4 egress pre-flight with the "cannot reach any IPv4 address / this is usually Docker" hint — but Docker is not installed and the container **does** have full IPv4 egress a few seconds later. The probe simply runs before the builder container has obtained its DHCP lease.

## Environment

- Hostinger KVM VPS, **1 vCPU / 4 GB RAM**, Ubuntu 24.04.4, kernel 6.8.0-137
- Fresh install via `curl -fsSL https://remote.futrx.com/get | sudo bash -s -- <host>` at commit `8b1b65b`
- LXD 5.21.6 (snap), no Docker, `iptables -P FORWARD ACCEPT`, `net.ipv4.ip_forward=1`

## What happens

```
16:49:10 [1/6] Downloading Ubuntu 24.04 and starting the builder...
16:49:35 [1/6] Downloading Ubuntu 24.04 and starting the builder finished in 26s
16:49:42 build failed: the builder container cannot reach any IPv4 address.

This is usually Docker: it sets the host's iptables FORWARD policy to DROP ...
```

Reproduced 2/2 runs. Meanwhile a manually launched `ubuntu:24.04` container on the same host, probed with the exact `ipv4EgressProbe` script at intervals after launch:

```
t+2s : bash: connect: Network is unreachable   (no IPv4 address yet)
t+5s : TCP-OK   (10.215.233.x/24 via DHCP, 1.1.1.1:443 reachable)
t+10s: TCP-OK
curl -4 https://github.com → HTTP/2 200
```

So on this box the DHCP lease arrives ~4–5 s after `LaunchContainer` returns, while `baseImageNetworkWarmup` is a fixed `3 * time.Second` and the probe fails on the first pass ("Network is unreachable" returns instantly for all three endpoints, so the whole probe finishes in well under a second).

## Why it matters

The error text explicitly points at Docker / `DOCKER-USER`, so an operator on a Docker-free host is sent chasing the wrong cause. The pre-flight was added to *avoid* misleading failures; on slow hosts it now produces one of its own.

## Suggested fix

Make the probe wait for IPv4 instead of testing once. Local patch I ran with successfully (`egress.go`):

```diff
-const ipv4EgressProbe = `for endpoint in 1.1.1.1 9.9.9.9 8.8.8.8; do
+const ipv4EgressProbe = `for attempt in $(seq 1 30); do
+  for endpoint in 1.1.1.1 9.9.9.9 8.8.8.8; do
     timeout 8 bash -c "exec 3<>/dev/tcp/${endpoint}/443" 2>/dev/null && exit 0
+  done
+  sleep 2
 done
 exit 1`
```

Alternatively, poll `lxc list` for an IPv4 address on the builder before probing (bounded, e.g. 60 s), and only then emit the Docker hint if the probe still fails. Either way the Docker hint would then only appear when IPv4 is genuinely blocked.

Happy to send a PR if you'd prefer one of these approaches.

mustafzidan74:
Fixes #39.

The pre-flight runs once, after a fixed 3s warmup. That assumes the builder has its DHCP lease by then, and on a 1-vCPU host it does not. The lease lands around 5s, so the probe tests a container with no address at all and the build stops with the Docker FORWARD-policy hint from egress.go. On a box with no Docker on it that message sends you straight into iptables looking for something that was never there.

It now retries for up to 45s at 3s intervals. A container that is only slow to come up gets the time it needs. One that genuinely has no IPv4 route still fails with the same message, just later, so the case the hint was written for is unchanged.

A cancelled build returns the context error instead of the network hint, since Ctrl-C is not a firewall problem.

The existing test only fed one failing probe response, which stopped being meaningful once the probe retries, so it now simulates a container that never gets IPv4. Three new tests cover the slow lease, the real block, and cancellation.

### DIFF
--- a/backend/internal/service/container/image/builder.go
+++ b/backend/internal/service/container/image/builder.go
@@ -185,8 +185,8 @@ func (b *Builder) Build(ctx context.Context, alias string) error {
 	}
 
 	// Fail fast on a container that can only reach IPv6 (see egress.go).
-	if _, err := b.runtime.ExecuteScript(bctx, baseImageBuilderName, ipv4EgressProbe); err != nil {
-		return errors.New(ipv4EgressHint)
+	if err := awaitIPv4Egress(bctx, b.runtime, baseImageBuilderName); err != nil {
+		return err
 	}
 
 	out, err = b.runBuildStage(2, "Installing system tools, Node.js, and agent CLIs", func() (string, error) {
--- a/backend/internal/service/container/image/builder_test.go
+++ b/backend/internal/service/container/image/builder_test.go
@@ -21,6 +21,9 @@ type recordingRuntime struct {
 	scriptResponses []runtimeResponse
 	stopResponse    runtimeResponse
 	publishResponse runtimeResponse
+	// scriptAlwaysErr fails every ExecuteScript, for the container that never
+	// comes up rather than the one that is merely slow to get a lease.
+	scriptAlwaysErr error
 }
 
 func (r *recordingRuntime) Available() bool {
@@ -46,6 +49,9 @@ func (r *recordingRuntime) ExecuteScript(_ context.Context, containerName, scrip
 		}
 	}
 	r.events = append(r.events, "script "+containerName+" "+script)
+	if r.scriptAlwaysErr != nil {
+		return "", r.scriptAlwaysErr
+	}
 	if responseIndex >= len(r.scriptResponses) {
 		return "", nil
 	}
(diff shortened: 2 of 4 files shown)


kingsoftheweb:
@mustafzidan74  this has been solved here: https://github.com/futrx-com/remote.futrx/pull/90

PR #91, opened 2026-08-29
Show subscription plan limits from what the CLIs report

mustafzidan74:
Follow-up to the conversation on #47 — you said you'd wait for the PR, so here it is. It does not touch the ledger.

### The gap

The usage page answers "what did this platform spend". It can't answer "how much of my Claude Max / ChatGPT plan is left", because the plan is spent from everywhere I work — this platform, my laptop, whatever else — and only the vendor knows the total. On my install that was the number I actually wanted before starting a long run.

### Where it comes from

Neither CLI has an endpoint for it, but both mention it while they work:

- claude emits a top-level `{"type":"rate_limit_event"}` line carrying one window at a time (`five_hour` / `seven_day`), with a status and an absolute reset time.
- codex puts `rate_limits` inside its `token_count` event, as a primary and a secondary window, each with a percentage used and a window length in minutes.

So the parsers keep what goes past. A small service holds the last reading per agent at `DATA_DIR/agent-quota.json`, and it renders above the ledger on Settings → Usage.

### Two things I was deliberate about

Both are the difference between a useful card and a confidently wrong one:

**A reading is a snapshot, and it says how old it is.** There is nothing to poll, so an idle platform is showing a number from whenever it last ran. Printing that as if it were current is how you plan a day's work against yesterday's figure.

**Absent is not zero.** Claude usually sends a status and no percentage — a bar drawn at 0% would claim the plan is untouched, which the CLI never said, so those rows get the word and no bar. A window name the platform doesn't recognise is dropped rather than filed under whichever window looks closest.

The service is built whether or not a store is present (without one, readings just don't survive a restart), and the persist is best effort off the run's critical path — losing a reading costs a stale card until the next run, while failing a turn over it would cost real work.

### Notes

- New endpoint: `GET /api/agent-quota`, any signed-in user. An empty list is a real answer, not an error: a platform nobody has run yet genuinely knows nothing, and the UI renders nothing rather than an empty gauge.
- A `token_count` line with no `rate_limits` block stays silent, so the ordinary per-turn counter doesn't file empty readings.
- Tests cover both parsers (including the only-one-window and no-rate-limits cases), the service (replace-same-window, unknown window dropped, survives a restart, stable order), and the display helpers (tone, countdown, age).
- I've had this running on my own install for a while; the screenshot-worthy part is just two rows with percentages and "resets in 2h 30m".

Happy to move it somewhere other than Settings → Usage if you'd rather it lived elsewhere.


mustafzidan74:
Thanks for the refactors — I've merged current `qa` in, so this is mergeable again.

One conflict, in `prompt/service.go`: both sides added a field to the `Service` struct — `quota` here, `startGate` on qa. Independent, so both are kept.

I went through your five commits before merging and they're faithful. Moving `normalizeQuotaWindow` into each adapter is better than where I had it — the wire vocabulary is Claude's, not the platform's, and the "unknown name stays unknown" rule survives with a comment saying why. `ResetsInSeconds` was my typo. And computing `barPercent` in state rather than in the component makes the important rule harder to lose by accident: a window with no percentage gets `null`, so nothing draws a bar at zero. All my behavioural tests still pass under the new layout, plus the two you added.

Separately, and not from this branch: `go test ./internal/service/auth/` doesn't build on `qa` right now. `service_test.go` line 85 is a stray ` B N` and gofmt refuses the file:

```
internal/service/auth/service_test.go:85:2: expected declaration, found B
```

It's at qa's tip and absent at the merge base, so it arrived on qa rather than here. I left it alone because it has nothing to do with plan limits — happy to send a one-line PR against qa if that's easier than picking it up yourselves.

PR #92, opened 2026-08-29
Ranked workspace search, a command palette, and find-in-chat

Ahmedwaleed22:
Replaces the sidebar's `title + cwd + model`.includes(query) filter with a ranked
search engine, puts a filter menu and a command palette on top of it, and adds
find-in-chat on Cmd/Ctrl+F.

## What's here

**Ranked search.** The old substring filter missed reordered words, separator
variants and typos, and could not rank one hit above another. The engine is pure
and layered as a DAG, driven by two registries so neither the engine nor any
component is edited to extend them:

- `SEARCH_FIELDS` declares each searchable field once — the doc's folded text,
  the ranking weights and the scoring loop all derive from it.
- `FACET_DEFINITIONS` declares each filter once — `FacetId`, the filter panel,
  the chips, the badge count and the per-option counts all derive from it.

Searching switches the sidebar from a filtered tree to a flat ranked list:
relevance ordering only means something across projects, and a tree hides the
best match inside a collapsed group.

**Filters.** Behind a discoverable button rather than a query syntax to learn.
Projects are multi-select; semantics are OR within a section, AND across
sections, and each option shows how many chats it would match given the *other*
active filters.

**Command palette.** Cmd/Ctrl+P or Cmd/Ctrl+K. Also gives search a home while
the sidebar is collapsed, which was previously impossible. It keeps its own
filter selection — see below.

**Find in chat.** Cmd/Ctrl+F opens a find bar scoped to the open chat. Enter and
Shift+Enter step and wrap, Escape closes. It searches the *rendered* thread
rather than the message model: that is what Cmd+F means everywhere else, and it
keeps the count honest about what is on screen — tool output and code blocks
included — instead of claiming matches in messages the list has not rendered.
Where older messages are still on the server, the bar says so rather than
quietly searching half a conversation.

Highlighting uses the CSS Custom Highlight API, so live Ranges are painted
without wrapping anything in `<mark>`: no reflow, nothing to unpick when the
query changes, and no risk of splitting an element the markdown renderer built.
Browsers without it (Safari before 17.2) lose the highlight only; counting,
navigation and scrolling still work.

## Two bugs found along the way

**Deleted projects appeared as filter options.** Chats keep their `projectId`
when a project is deleted, and both the facet registry and the sidebar tree took
that dangling id at face value — so a bare hex id showed up as a project nobody
created. Worse, `bucketChatsByProject` filed those chats under a project id that
is never rendered, so they vanished from the sidebar entirely. Both now treat an
unresolved id as no project, which puts the chats under Unassigned where they
can be found.

⚠️ **The backend cause is untouched and still live.** `project.Service.Delete`
cleans up the container, secrets and access entries but leaves the project's
chats behind, so every project deletion strands its chats. Fixing it means
deciding whether deleting a project should delete its chats or orphan them
deliberately, plus a cleanup of the ones already on disk — both destroy user
data, so I left it for a separate call.

**Opening the filter menu could scroll the sidebar's header away.** Selecting an
option made the sidebar jump: its logo row and search input slid out of sight
with no scrollbar anywhere to undo it. `.codex-window-frame` set
`overflow: hidden`, which still makes the card a scroll container — it only
hides the scrollbar. The menu is absolutely positioned inside that card and
taller than the space beneath the search bar, so the card's `scrollHeight`
(869px) ran past its `clientHeight` (725px); focusing the `sr-only` checkbox
behind an option made the browser scroll the card to reveal it.

`overflow: clip` clips identically but is not a scroll container. `hidden` stays
ahead of it as the fallback for browsers without `clip`. This covers every
popover in a window card, not just this one — the composer's dropdowns and the
skill picker sit in the same kind of card.

## Note for review

The palette and the sidebar deliberately do **not** share a filter selection.
They did at first, which meant narrowing the palette to a project silently
re-scoped the sidebar behind it, and — because that state persisted — outlived
the session. Persistence is now a `SearchPreferences` seam: the sidebar's search
is stored, the palette starts from defaults and saves nothing.

## Testing

- 128 unit tests pass (`node --test`), `tsc` and `vite build` clean.
- Search engine, facets, date ranges, sort and storage are covered by
  `search.test.ts`.
- DOM-dependent behaviour was verified by driving the real components in
  headless Chrome: 20 checks over find (matches split across elements, case and
  accent folding, `data-find-skip` exclusion, document order, counting, both
  highlight layers, wrap-around, scroll-to-match, empty state, Escape) and
  before/after measurements of the sidebar scroll bug (`scrollTop` 144 → 0,
  search input `top` −71 → 73).

## Known gaps

- Pressing Cmd+F while the command palette is open opens the find bar behind it.
  Wiring the guard means giving `ChatContainer` a dependency on palette state.
- Find covers loaded messages only, as described above.


frontend/src/state/hooks/chat/useChatFind.ts
```diff
@@ -0,0 +1,145 @@
+import { useCallback, useEffect, useState } from "preact/hooks";
+import type { RefObject } from "preact";
+import { clearHighlight, paintHighlight } from "../../../services/platform/textHighlight.ts";
+import { findRanges } from "../../search/domTextSearch";
+import { isFindShortcut } from "../../search/searchShortcuts";
+
+/** Keep the match this far from the scroller's edges when revealing it. */
+const REVEAL_MARGIN = 80;
```
kingsoftheweb: @Ahmedwaleed22  move that to the configs/ folder, as this could be reused by another component in the future

frontend/src/state/hooks/chat/useChatFind.ts
```diff
@@ -0,0 +1,145 @@
...
+
+// Two layers so the current match reads differently from the rest. The current
+// one is painted separately rather than held out of `ALL`, so its rule wins by
+// being registered second. Styled in index.css as `::highlight(...)`.
+const ALL_MATCHES = "chat-find";
```
kingsoftheweb: @Ahmedwaleed22  you can prefix this with SEARCH_* and add it to the configs/

frontend/src/state/hooks/chat/useChatFind.ts
```diff
@@ -0,0 +1,145 @@
...
+const CURRENT_MATCH = "chat-find-current";
+
+function paintMatches(all: readonly Range[], current: Range | null): void {
```
kingsoftheweb: @Ahmedwaleed22  move to services

frontend/src/state/search/dateRange.ts
```diff
@@ -0,0 +1,136 @@
...
+// Ranges resolve in the user's local timezone: someone picking "Today" means
+// their calendar day, not UTC's. `now` is always injected so the resolver stays
+// pure and testable.
+
+// The preset and field vocabularies are declared once, as the option tables the
+// menu renders; the id unions and the lists storage validates against are
+// derived from them, so a new preset cannot be added to one and missed in the
+// other.
+
+export const DATE_PRESET_OPTIONS = [
```
kingsoftheweb: @Ahmedwaleed22  move to configs

frontend/src/state/search/domTextSearch.ts
```diff
@@ -0,0 +1,73 @@
+import { fold } from "./textMatch";
```
kingsoftheweb: @Ahmedwaleed22  all the search states, need to be converted to stores to follow the new pattern of using zustand in the stores, while keeping the hooks as the abstraction layer for the UI components to talk to

kingsoftheweb (changes requested):

frontend/src/state/hooks/chat/useChatFind.ts
Ahmedwaleed22: do you mean `const REVEAL_MARGIN = 80;` ?

PR #97, opened 2026-08-30
fix(lifecycle): raise container launch timeout to 10 minutes

### LINKED ISSUE #96: Workspace container creation fails with signal: killed when lxc init exceeds 90s
Running infra/upgrade-workspaces.sh fails on every project container with:

FAIL project_name: lxc init: signal: killed; output: Creating project_name
Retrieving image: Unpacking image: 81% (20.44MB/s)
rollback launched container: lxc delete: exit status 1; output: Error: Failed deleting instance "project_name" in project "default": Failed creating instance delete operation: Instance is busy running a "create" operation
workspace upgrade: 0 upgraded, 0 busy skipped, 2 failed

Cause: launchTimeout in internal/integration/containers/lifecycle/runtime.go is 90 seconds. It bounds the lxc init call on line 49. Unpacking the base image (Ubuntu 24.04 plus Node 22 plus the four agent CLIs) takes several minutes at typical disk throughput on a dir storage backend. When the context expires, RunWithTimeout sends SIGKILL and the failure surfaces as signal: killed.

Not a resource problem. The host had 9.7GB available memory, 8GB of unused swap, and 164GB free disk. dmesg shows no OOM kill.

Two follow-on effects make this worse than a plain timeout:

lxc init is only the client. Killing it does not stop the daemon, which keeps unpacking server side. Rollback then fails with "Instance is busy running a create operation", so the old container is already deleted and no new one is usable. Workspace files survive because they are bind-mounted from the host, but the containers are left bare with devices: {} and no futrx-workspace profile.
On a retry, Go issued lxc start while a create operation was still running on the daemon, producing Failed to create instance start operation: Instance is busy running a "create" operation.

Suggested fix: raise launchTimeout to 10 minutes, matching the existing 5 minute migrateTimeout and deleteTimeout in the same block.

Worth considering separately:

Detect that the daemon still holds a create operation after a killed client, and wait or reconcile instead of proceeding to start and rollback.
A btrfs or zfs storage pool unpacks the image once and clones per container, which removes the multi-minute unpack from the create path entirely.

Ahmedwaleed22:
launchTimeout was 90 seconds, which bounds the lxc init call in runtime.go:49. That is not enough to unpack the base image, so every workspace create was SIGKILLed partway through and upgrade-workspaces.sh failed with 0 upgraded, 2 failed.

Raises it to 10 minutes.

launchTimeout has exactly one caller, so nothing else is affected. The value is in line with migrateTimeout and deleteTimeout, both already 5 minutes.

Fixes #96

### DIFF
--- a/backend/internal/integration/containers/lifecycle/runtime.go
+++ b/backend/internal/integration/containers/lifecycle/runtime.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	launchTimeout  = 90 * time.Second
+	launchTimeout  = 10 * time.Minute
 	migrateTimeout = 5 * time.Minute
 	startTimeout   = 30 * time.Second
 	stopTimeout    = 30 * time.Second
--- a/backend/internal/integration/containers/lifecycle/runtime.go
+++ b/backend/internal/integration/containers/lifecycle/runtime.go
@@ -15,6 +15,7 @@ import (
 )
 
 const (
+	// TODO: We will need to investigate how to hook into a callback instead of increasing the timeout
 	launchTimeout  = 10 * time.Minute
 	migrateTimeout = 5 * time.Minute
 	startTimeout   = 30 * time.Second


backend/internal/integration/containers/lifecycle/runtime.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	launchTimeout  = 90 * time.Second
+	launchTimeout  = 10 * time.Minute
```
kingsoftheweb, line 19: @Ahmedwaleed22  add the todo here

kingsoftheweb (changes requested):

PR #100, opened 2026-08-31
feat: add OpenCode agent support (host + project scopes)

mohammedbakri123:
## Summary

Adds OpenCode (`opencode-ai` CLI) as an agent provider, following the same provider-owned factory architecture as Claude/Codex/Kimi/Antigravity.

**Covers Phases 1–3:** host + project-container scopes, skills instruction triggers, scheduled tools, and browser asset migration (matching the Kimi integration surface).

## What's included

- New provider ID `opencode` + version pin `OPENCODE_VERSION=1.18.25` in the single versions manifest
- New package `backend/internal/integration/agents/opencode/`:
  - Provisioning profile: npm install of `opencode-ai`, `~/.local/share/opencode/auth.json` credential file (XDG-aware)
  - Auth: managed device flow via `opencode auth login`
  - Parser: `opencode run --format json` JSONL to normalized agent events (text, reasoning, tool start/complete, usage, session updates)
  - Command: `--model provider/model`, `--session` resume, native `--session <id> --fork`, `--agent plan` for Plan mode
  - Capabilities: live model list from `opencode models`, fallback to Auto
- Registered in the agent catalog (host + project scopes); appears in Settings → Agents
- Frontend: `opencodeSessionId` tracked in chat metadata/session events
- Infra: host-agent install plan includes opencode (`infra/tests/host-agent-install-test.sh` updated and passing)
- Docs: ARCHITECTURE.md + user guide "OpenCode sign-in and behavior" section

## Validation done

- `go test ./internal/...` all green; `go vet` / gofmt clean
- Parser tests use **real captured output** from live runs of both CLI **1.15.13** and the pinned **1.18.25** (text-only and tool-using turns)
- Frontend projector tests pass
- `infra/tests/host-agent-install-test.sh` passes with opencode in the plan

## Known caveats (for maintainers / testers)

1. **Not yet run on a deployed QA server by the author** — unit-tested only. Suggested QA: `bash infra/qa/update.sh <ref>` (container scopes changed), then one real opencode chat on the host (`opencode auth login` first) and one in a project container.
2. Parser is verified against the pinned 1.18.25 wire format; 1.15.13 fixtures are kept too, since older versions sometimes end a run without the terminating `step_finish(reason=stop)` — the provider synthesizes `run.completed` on clean exit in that case.

## Not yet included

- Per-run browser MCP tool wiring (OpenCode gets browser asset migration only, like Kimi — per-run MCP is Claude/Codex only)

## How to smoke-test

```bash
git checkout feat/opencode-agent
cd backend && go test ./internal/...
# on a host with opencode installed and authenticated:
# deploy, then: Settings → Agents → OpenCode → host-scope chat
```


ActArtech:
Hi @mohammedbakri123 — great work on this PR. I'd love to see OpenCode land in remote.futrx, so I picked it up and finished the path to mergeable on top of your branch (all credit for the original integration is yours). What was done:

- **Rebased your 11 commits onto current `qa`** (it had moved ~395 files since your base). Conflicts resolved in `versions.env` (kept qa's newer CLI pins), `agents.go` (MiniMax and OpenCode coexist), `chat.ts` (qa's new event types + your `opencodeSessionId`), `ARCHITECTURE.md`, and the browser-MCP docs caveat.
- **Fixed a stray token** in `backend/internal/service/auth/service_test.go` that broke `go vet`/`go test` compilation (CI doesn't run `go test`, so it was easy to miss).
- **Fixed the `agents_test.go` order expectations** that git auto-merged incorrectly — MiniMax landed on qa after your base, so project profiles now include it, while host profiles correctly don't (MiniMax is project-scope only).
- **Bumped the pin 1.18.25 → 1.18.29** and validated the JSONL wire format against a real 1.18.29 capture: identical envelope/part shapes; 1.18.29 only adds an optional `time` object on text parts, which the parser already ignores. Added a test pinning a real v1.18.29 capture.
- **DCO sign-offs on all commits** (per CONTRIBUTING), your authorship preserved on your 11.

Verification on a clean Ubuntu 24.04 box: `go build`/`go vet`/`go test` 77/77 packages, `gofmt` clean, frontend `tsc`+vite build clean with 244/244 tests, `infra` host-agent-install-test and release-version-test pass (opencode → `npm opencode-ai` @ 1.18.29). The backend also boots and serves on a live host, and a real `opencode run --format json` stream parses end-to-end.

The finished branch (16 commits: your 11 + 5 fixes) is on my fork:
https://github.com/ActArtech/remote.futrx/tree/feat/opencode-agent

Happy to hand it over however you and the maintainers prefer: you're welcome to push this branch to this PR, or if you'd rather not continue with it, I can open a replacement PR from my fork that cites #100. Your call — thanks again for the solid foundation.

mohammedbakri123:
Thanks so much @ActArtech! Appreciate the clean rebase and the 1.18.29 bump.
 Just updated the PR with your branch. Had to patch a couple of CI runner issues as well, but all checks are green and passing now. Thanks again!

Mohamed00hosen (approved):

Mohamed00hosen (approved):

Mohamed00hosen (changes requested):
Thanks @mohammedbakri123 and @ActArtech for the work here — the structure follows the Kimi pattern well and the fixtures from real captures are appreciated. I went through the diff against the pinned opencode v1.18.29 source (`packages/opencode/src/cli/cmd/run.ts`) and found a few things that need to be fixed before this can land, plus some items that need live verification since the branch hasn't been run on a QA host yet.

## Must fix

### 1. Tool errors are mapped to `tool.started` and never complete
`opencode/parser.go` → `parseToolUse` treats any status other than `completed` as `EventToolStarted`. In v1.18.29 `--format json` only emits `tool_use` when the state is `completed` **or** `error`:

```ts
if (part.type === "tool" && (part.state.status === "completed" || part.state.status === "error")) {
  if (emit("tool_use", { part })) continue
```

So a failed tool call shows up as "running" in the UI forever. `state.error` is also not decoded in `wireToolState`.

- Map `status == "error"` to `EventToolCompleted` with `IsError = true` and `Output = state.error`.
- Add an `Error` field to `wireToolState`:
  ```go
  Error string `json:"error"`
  ```
- `fixtureToolStart` (status `pending`) tests an input the CLI never produces in JSON mode — replace it with a real `error`-state fixture.

### 2. `error` event shape is wrong
The parser reads `part.text` / `part.error`, but opencode emits the error at the top level:

```ts
emit("error", { error: props.error })
// → {"type":"error","timestamp":…,"sessionID":"…","error":{"name":"…","data":{"message":"…"}}}
```

Result: the fallback dumps the whole raw JSON line as the user-facing error message. Please add a top-level `Error json.RawMessage` to `wireEvent` and extract `error.data.message`, falling back to `error.name`.

### 3. `opencodeSessionId` on the frontend is dead code
`backend/internal/service/chat/model.go` only mirrors the four legacy fields (`claude/codex/kimi/antigravitySessionId`) and explicitly describes them as a compatibility window. The backend will never emit `opencodeSessionId`; the generic `sessions["opencode"]` map already carries it (same as MiniMax). Please drop the additions in `frontend/src/models/chat.ts`, `workspaceDataProjector.ts`, and the projector test.

## Needs a decision / live verification

### 4. Prompt passed as positional gets mangled by opencode
In v1.18.29:

```ts
let message = [...args.message, ...(args["--"] || [])]
  .map((arg) => (arg.includes(" ") ? `"${arg.replace(/"/g, '\\"')}"` : arg))
  .join(" ")
```

Any prompt containing a space reaches the model wrapped in `"…"` with inner quotes escaped. Piped stdin is **not** transformed (`resolveRunInput`). A prompt starting with `-` is also parsed as a yargs flag, and the prompt is visible in `ps`.

Suggestion: write the prompt to `cmd.Stdin` instead of appending it to `args`. opencode reads stdin to EOF when it's not a TTY, and `RunProcess` currently leaves stdin as `/dev/null`, so switching is safe. Please add a test that the prompt is not in `cmd.Args`.

### 5. Permissions are auto-rejected in non-interactive `opencode run`
```ts
if (event.type === "permission.asked") { … auto ? reply("once") : reply("reject") }
```

Without `--auto`, any tool that hits a permission rule fails silently depending on the user's opencode config. We need to decide the policy here — likely pass `--auto` in the project-container scope (isolated), and match whatever Claude/Codex do inside containers. At minimum document the behaviour.

### 6. Managed device auth flow probably doesn't work headless
`opencode auth login` is an interactive clack prompt (provider select → method select). Driven as a subprocess without a TTY the prompts cancel/exit, so `DeviceService` will likely never see a URL. The user guide also says "run `opencode auth login` on the host" (manual), which contradicts `AuthManagedDevice` in the descriptor.

Please verify this on a real host. If it fails, switch to the Antigravity-style manual instructions rather than shipping a broken UI flow.

### 7. Host capability probe hardcodes `HOME=/root` / `XDG_DATA_HOME=/root/.local/share`
`capabilities.go` passes these unconditionally; `mergeEnvironment` overrides the process env, so on a host with a different `HOME`/`XDG_DATA_HOME` the probe looks in the wrong place while `auth.go`/`profile.go` use `hostOpenCodeData()`. Kimi handles this correctly (`hostKimiHome()` when `req.ContainerName == ""`). Same inconsistency in `buildCmd`: the host path doesn't set `XDG_DATA_HOME`, while `opencodeEnv` does for login.

### 8. Unrelated change in `infra/install.sh`
The root-gate reorder has nothing to do with OpenCode. Please move it to its own PR (like #162) so this one stays focused.

## Nits

- `profile.go`: `missingAuthJSONFormat` is unused.
- `command.go`: `normalizeModel` is just `strings.TrimSpace`; inline it.
- `agents_test.go`: the Antigravity check via re-slicing `hostProfiles` is hard to read — look it up by ID instead.
- `capabilities_test.go`: the `len(caps.Models) == 0` check is redundant after asserting `== 4`.
- Text parts are only emitted once `part.time.end` is set, so there's no token-level streaming — fine, but worth a comment in the parser doc.
- The parser comment about the CLI "not always emitting `step_finish`" applies to 1.15.x; 1.18.29 emits it reliably. Keep the fallback but update the wording.

## Process
My earlier approvals predate the last force-push (Sep 8 19:41), so please treat them as stale — I'll re-review once the above is addressed. Once fixed, one real host-scope chat and one project-container chat on QA (`bash infra/qa/update.sh <ref>`) with a tool call that fails and one that hits a permission prompt would give us good coverage of items 1, 5 and 6.

Mohamed00hosen (approved):
Re-reviewed `8a923d9` — all eight items and the nits are addressed. Nice turnaround.

Verified in the diff:
- Parser: `error` tool state → `tool.completed` + `IsError`, top-level `error` payload decoded with `data.message` → `name` → raw fallback, both with tests.
- Prompt goes through `cmd.Stdin` on host and container; test asserts it's absent from `cmd.Args`.
- `--auto` only in project-container default mode; plan mode and host unchanged; documented.
- Auth switched to `AuthExternal`; `auth.go` and the device flow are gone.
- Host capability probe and `buildCmd` now derive `XDG_DATA_HOME` from `hostOpenCodeData()` without overriding `HOME`.
- Frontend `opencodeSessionId` and the `install.sh` change are dropped.

Two things to confirm on the QA deploy, since this hasn't run live yet:
1. `lxc exec` forwards stdin to `opencode run` in the container path — it should in non-interactive mode, but a real project-container chat is the only way to be sure.
2. One tool call that fails inside a container renders as a completed/errored tool rather than a stuck spinner.

Approving; the `qa` branch is the right gate for those.

Mohamed00hosen:
@kingsoftheweb — this one is ready for a maintainer look. Summary for a quick pass:

- Adds OpenCode (`opencode-ai` 1.18.29) as a compiled-in agent module following the Kimi pattern: host + project-container scopes, skills instructions, scheduled tools, browser asset migration.
- I reviewed the diff against the pinned opencode CLI source and requested changes on parser/error handling, prompt passing, permissions, auth flow and env handling; @mohammedbakri123 addressed all of them in `8a923d9` and I've re-approved.
- Auth is `AuthExternal` (same as Antigravity), so there's no new managed login UI to break.
- Not yet run on a deployed QA host by anyone. Two things worth watching on the `qa` deploy: `lxc exec` forwarding the prompt via stdin in project containers, and a failed tool call rendering as errored rather than stuck.

Note that my approval is advisory — I don't have write access here, so the merge call is yours.

PR #103, opened 2026-08-31
fix(auth): gate first-boot credential claim behind a server-printed token

### LINKED ISSUE #65: Move first-time user credential setup to server terminal
## Problem

The first-time user credential setup is currently available through the public web page.

This is an existing issue in the current implementation and can allow anyone who can access the page to set the credential before the actual owner.

## Expected

The first-time credential setup should be done from the **server terminal** instead of the public web page.

The web app should only show whether the credential is configured and should not allow users to claim or initialize it.

## Requirements

* Move first-time credential setup to the server terminal.
* Remove the public setup/claim flow.
* Keep the credential stored securely on the server.
* Show the credential status in the web app.
* Add a test to prevent the public flow from initializing the credential.

## Note

This issue describes an **existing problem in the current implementation**.

I am opening this issue first and will submit a PR to address and fix it.

AbdallahMohamedDotnet:
Closes #65

## What

First-time local-admin setup is now gated on a one-time token that is generated on the server and printed only to its terminal. `POST /auth/local/claim` refuses any first-boot claim that does not present it.

## Why

On a genuinely first boot the user directory is empty, so no administrator exists to authorise a claim — and the endpoint handed ownership of the server to whoever reached the page first. Once an administrator *does* exist the claim was already authorised against them (`local_admin_authenticator.go`), so that branch was never the exposed one. Only the empty-directory branch is newly gated; the administrator-authorised path is untouched, since demanding a terminal token from a caller who is already authenticated would be friction without a security gain.

## Design note for reviewers

Issue #65 asks to **remove** the public setup flow. This keeps the web form but makes it inert without a terminal-issued token — same security property, no CLI-only UX regression. Flagging explicitly in case you would prefer literal removal; happy to change it to `Refs #65` if you would rather keep the issue open pending that decision.

**The token is delivered in the URL fragment, not a query string.** Browsers never transmit a fragment, so it cannot reach a reverse-proxy access log. A `?token=` form would have been logged by Caddy on the very first page load, before any form was submitted.

**The token is spent only after the claim fully succeeds.** A weak password, an invalid email, or a rolled-back bootstrap all leave it usable — burning it on a recoverable failure would strand the operator with neither a credential nor a way back in.

## Also fixed (independent, first commit)

The claim rate limit was keyed on the attacker-supplied **leftmost** `X-Forwarded-For` entry. Caddy appends rather than replaces, so rotating the header gave a fresh bucket every request and defeated the five-failure lockout entirely. Now keyed on the rightmost hop — the one our own proxy appended. Keying on `RemoteAddr` was not an option: Caddy proxies from loopback, so every request would share one global bucket and five failures from anyone would lock out every user.

## Operator-facing

- Unclaimed server prints the setup link at startup and **rotates on every restart**, so a token that leaked beforehand is already dead.
- `remote setup-token` reissues from the host if the link is lost or expired.
- A configured server prints nothing, so a live token never lands in a running system's log.
- README, ARCHITECTURE, troubleshooting, and known-limitations updated. The troubleshooting entry previously read *"the claim endpoint is public only while the server is unclaimed"* — that was the vulnerability, stated as documentation.

## Verification

Backend and frontend suites green against the `main` baseline. Verified end-to-end against a running server, not just in unit tests: tokenless/wrong/expired/replayed tokens rejected with no credential written; a valid token accepted exactly once; the token appearing exactly once in the log, never in a response body, and never in plaintext on disk (only its SHA-256, mode `0600`); six concurrent claims on one token producing exactly one winner; the rate limit holding against a rotated forwarded header.

Four rounds of adversarial QA ran against this branch. Three defects were found and fixed in-branch — notably two that were *introduced by the fix for the previous one*, both small additive-looking changes with non-obvious startup side effects (a corrupt `users.json` preventing boot; the reissue command creating a root-owned `session.key` that the service then could not read).

## Known gaps

- The frontend fragment read/strip has unit coverage of its policy and the shipped bundle was inspected, but there is no browser-level test.
- Token expiry is covered with an injected clock, not a real 30-minute wait.
- The rate limit remains per-IP, so a distributed attacker is not slowed — pre-existing, shared with `/auth/local/login`.
- Rolling back to a release without this change reopens the claim on an unclaimed server; worth a release-note line.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


backend/internal/stores/fileauth/store.go
```diff
@@ -148,6 +148,65 @@ func (s *Store) DeleteLocalAdmin(ctx context.Context, expected serviceauth.Local
...
+	if record.Hash == "" {
+		return errors.New("setup token record is incomplete")
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.writeJSONLocked("setup-token.json", record)
+}
+
+func (s *Store) DeleteSetupToken(ctx context.Context) error {
```
kingsoftheweb: @AbdallahMohamedDotnet  is their any caller for this?

backend/internal/service/auth/setup_token.go
```diff
@@ -0,0 +1,124 @@
...
+	"fmt"
+	"time"
+)
+
+const (
+	// setupTokenBytes is the entropy behind the printed token. At 32 bytes the
+	// token is not guessable, so the rate limit in front of the claim endpoint
+	// is defence in depth rather than the thing standing between an attacker
+	// and the credential.
+	setupTokenBytes = 32
```
kingsoftheweb, line 20: @AbdallahMohamedDotnet  configs/constants need to move to configs/

frontend/src/ui/auth/LoginScreen.tsx
```diff
@@ -29,15 +29,24 @@ export function LoginScreen({
...
+  const title = awaitingSetupToken
+    ? "Finish setup from the server terminal"
+    : mode === "claim"
     ? "Create your admin account"
     : mode === "legacy-setup"
       ? "Secure your admin account"
       : "Sign in";
-  const description = mode === "claim"
+  const description = awaitingSetupToken
+    ? "This server has no administrator yet. Open the setup link printed in the server log, or run \u0060remote setup-token\u0060 on the host to print a new one."
```
kingsoftheweb: @AbdallahMohamedDotnet  what if the user forgot to use the correct url that has the token? is there a way for them to re-initiate it?

README.md
```diff
@@ -147,8 +147,14 @@ Replace `remote.example.com` with the hostname you set up above. The installer d
 
 ### 3. Create your first project
 
-1. Visit `https://remote.example.com`.
-2. Create the administrator account.
+1. On the server, read the setup link printed when Remote started:
+   `journalctl -u remote --since "-10 min" | grep -A2 "first-time setup"`.
```
kingsoftheweb, line 152: @AbdallahMohamedDotnet  do we really need that ?

backend/internal/stores/fileauth/store.go
AbdallahMohamedDotnet: There is no production caller for DeleteSetupToken at the moment. The production flow uses SaveSetupToken to overwrite/rotate the token, while SetupTokenGuard.Consume marks it as consumed. It’s currently only kept for tests/future hard-delete support. If we don’t need that, I can remove it from the store interface.

backend/internal/service/auth/setup_token.go
AbdallahMohamedDotnet: i'm fix it on this commit 
commit id ==> b358197171abcfce4929c2ef97ad6edf556474e4
refactor(auth): move setup token TTL into config

Setup token TTL was a hardcoded package constant wired directly into

the auth service instead of flowing from the composition root like

the other account security policies (PendingLoginTTL, EnrollmentTTL,

...). Add SetupTokenTTL to auth.Options/service.AuthOptions/

config.AuthOptions with a SETUP_TOKEN_TTL env override, thread it

through cmd/remote's server and CLI wiring, and cover it with tests

for the config default/override, Options validation, and end-to-end

wiring into the service.

frontend/src/ui/auth/LoginScreen.tsx
AbdallahMohamedDotnet: i'm fixed it on this commit 
commit id ==> 3b81436ea87df9a8d9648d8fa01b66d56d971943
fix(auth): improve setup instructions for unconfigured servers

README.md
AbdallahMohamedDotnet: i'm fixed it 
commit id ==> 1b91df32e3615a4a347f20c58a3d9824771df2f6
fix(docs): clarify setup instructions for first-time users

frontend/src/services/projects/QRGenerator.js
```diff
@@ -156,7 +156,7 @@ function addErrorCorrection(data, version) {
 function alignmentPositions(version) {
   if (version === 1) return [];
   const count = Math.floor(version / 7) + 2;
-  const step = version === 32 ? 26 : Math.ceil((version * 4 + count * 2 + 1) / (count * 2 - 2)) * 2;
+  const step = version === 32 ? 26 : Math.floor((version * 4 + count * 2 + 1) / (count * 2 - 2)) * 2;
```
kingsoftheweb, line 159: @AbdallahMohamedDotnet  is this to be expected?

Merge origin/qa into this branch to see if it still shows up

backend/cmd/upgrade-workspaces/main.go
```diff
@@ -14,6 +14,7 @@ import (
```
kingsoftheweb, line 14: @AbdallahMohamedDotnet  why we are making updates here ?

backend/internal/stores/fileauth/store.go
kingsoftheweb: I am not seeing any value in keeping it tbh. and if we need it in the future we can just add it. you can remove it now.

AbdallahMohamedDotnet: done i'm removed this function in commit ==> dc0a52ea8e60f5792addacec8b809b42122d5a1e

frontend/src/services/projects/QRGenerator.js
AbdallahMohamedDotnet: done

backend/cmd/upgrade-workspaces/main.go
AbdallahMohamedDotnet: This file's changes aren't part of your setup-token-gate work — they came in through the merge from origin/qa

PR #117, opened 2026-09-01
Qa

AbdallahMohamedDotnet:


AbdallahMohamedDotnet:
Splitting out the QR code fixes into a dedicated PR: #120. Closing this one since its QR content is superseded there; the setup-token-gate work should be re-submitted separately if still needed.

PR #132, opened 2026-09-02
fix(chat): prevent long text from overflowing the viewport

### LINKED ISSUE #62: Text overflows the viewport on mobile and desktop
### Problem
Long text does not wrap and spills past the edge of the screen. This happens on both mobile and desktop, so it is not just a small-screen breakpoint issue. Users have to scroll horizontally to read the content, and part of the text is cut off in the preview.

### Screenshot
Text overflowing the container on desktop

### Expected behavior
Text should wrap inside its container and stay within the viewport at all screen sizes.

### Likely cause
The text container is missing wrapping rules (for example `overflow-wrap: anywhere` or `word-break: break-word`) and/or has a fixed width that exceeds the viewport.

### Proposed fix
I am happy to open a PR for this. The change should be small: add proper wrapping to the affected text element and make sure the container uses `max-width: 100%`. Let me know if you would prefer a different approach.

FekyBaz:
Closes #62.

## What
Long unbroken runs (bare URLs, inline code, file paths) spilled past the viewport on mobile and desktop. This adds defense-in-depth wrapping so chat content always stays inside its column:

- scroll column clips horizontal overflow and can shrink (\MessageList\: \overflow-x-hidden\, \min-w-0\)
- flex parents/items can shrink (\AssistantMessage\, \UserMessage\, \ToolShell\ label: \min-w-0\)
- prose, thinking, user bubble, paragraphs, headings, blockquotes, table cells wrap anywhere (\[overflow-wrap:anywhere]\ + \reak-words\)
- inline \code\ and links break anywhere (\reak-all [overflow-wrap:anywhere]\)

Fenced code blocks and tables keep their existing internal \overflow-x-auto\ scrolling - untouched.

## Why
Matches the approach suggested in the issue (\overflow-wrap: anywhere\, \max-width: 100%\) and the existing codebase pattern already used in \StreamingText\ and the error box.

## Verified locally
- frontend \
pm test\ -> 163/163 pass
- frontend \
pm run build\ (tsc -b + vite) succeeds


kingsoftheweb:
@copilot resolve the merge conflicts in this pull request

PR #139, opened 2026-09-03
feat(chat): add slash command palette to the composer

### LINKED ISSUE #57: Typing `/` in the chat composer doesn't show available commands
## Summary

The chat composer's placeholder says "**/ for commands**", but typing `/`
does nothing — no command menu appears. The only way to pick a
command/skill today is the separate **Skills** dropdown button, which isn't
discoverable and doesn't match the `/`-to-search pattern people expect from
chat/command interfaces.

## Steps to reproduce

1. Open any chat.
2. Click into the composer input.
3. Type `/`.

## Expected

A filterable palette pops up listing the available commands for the current
provider/project — type to filter, arrow keys to move, Enter/Tab to select,
Esc to dismiss. Selecting one attaches it as a command/skill.

## Actual

Nothing happens. `/` is treated as a plain character. The placeholder text
("`/` for commands") advertises a feature that isn't wired up. Commands are
only reachable via the **Skills** dropdown.

## Impact

- The composer's own placeholder is misleading.
- Commands/skills are hard to discover and slow to reach.
- Diverges from the standard slash-command UX (Slack, Discord, Linear, etc.).

## Notes

- Command data already exists via `useAvailableSkills` / the Skills dropdown;
  this is a missing composer affordance, not missing data.
- Environment: web app (chat composer).

FekyBaz:
## Summary

Typing \/\ in the chat composer now opens a filterable palette of available commands for the current provider and project, matching the placeholder text that already advertised this feature.

## What changed

- **\commandQuery.ts\** — pure helpers for trigger detection (\commandQuery\) and command filtering (\ilterCommands\), with unit tests pinning behavior (trigger cases, command-prefix vs broad search, case-insensitivity).
- **\CommandPalette.tsx\** — viewport-pinned listbox (mirrors \SkillPicker\ so it isn't clipped by the thread column) with arrow-key navigation, Enter/Tab selection, Esc dismissal, and live filtering as you type.
- **\PromptTextarea.tsx\ / \ChatComposer.tsx\** — route keyboard events to the palette when open and attach a selected command exactly like the existing Skills dropdown.

## Verification

- \
pm run build\ (\	sc -b && vite build\) passes
- \
pm test\ — all 169 tests pass (6 new)

Closes #57


kingsoftheweb:
@FekyBaz  is this different to https://github.com/futrx-com/remote.futrx/pull/56 ?

FekyBaz:
@kingsoftheweb Yes — it’s essentially the same feature as #56.

#56 was already merged and introduced the inline / command palette. #139 is
a rework of that implementation: it keeps the same user-facing behavior,
but simplifies the architecture by moving the trigger/filter logic into
pure helpers and making the palette self-contained in ChatComposer, with
additional unit tests.

So #139 isn’t intended as a separate feature; it’s a cleaner/reworked
implementation of the slash-command palette from #56.

kingsoftheweb:
@FekyBaz  sounds good. can you fix conflicts so I merge it?

PR #140, opened 2026-09-03
Feat/computer use flow map

Amine345-v:


kingsoftheweb:
@Amine345-v  I like this . thanks for the PR.

What are your thoughts on finding a way to integrate this, to give access to a personal computer from the remote server using tailscale or whatever type of proxy?

I haven't looked into it. And since you already spent some time on this, you might have thoughts about it.

This could be extremely beneficial for such cases when someone needs certain apps on their own personal computers that wont run on a linux env, and a computer use from remote can take over. This could completely eliminate the need of using any desktop application on any personal computer's again.

PR #146, opened 2026-09-03
feat(infra): support custom Caddy snippets for DNS-01 and private hosts (#27)

### LINKED ISSUE #27: Support DNS-01 ACME challenges for hosts without public inbound reachability
### Summary

Remote obtains certificates through Caddy's on-demand TLS, which relies on the HTTP-01 / TLS-ALPN-01 challenges. Both require Let's Encrypt to reach the host from the public internet.

That rules out a class of self-hosted deployments where the machine is *intentionally* not publicly reachable: Tailscale/WireGuard tailnets, private LANs, and home servers behind CGNAT. Since Remote is aimed squarely at self-hosting, this reads more like a gap than an edge case.

### Current behaviour

`infra/templates/Caddyfile.tmpl` sets up:

- a global `on_demand_tls` block with `ask http://127.0.0.1:${SERVICE_PORT}/internal/tls-ask`
- `tls { on_demand }` on `*.code.${HOSTNAME}` and `*.dev.${HOSTNAME}`
- no ACME DNS provider

`infra/steps/03-caddy.sh` re-renders `/etc/caddy/Caddyfile` from that template, so editing the generated file is not durable across installs or updates — the only way to keep a DNS-01 configuration today is to maintain a fork.

The installer already anticipates unusual DNS topologies with `--skip-dns-check` ("Cloudflare proxy / tunnels / etc"), but there is no matching escape hatch for certificate *issuance*.

### Proposal

Optional DNS-01 support, off by default:

1. Two optional variables, e.g. `ACME_DNS_PROVIDER` and `ACME_DNS_TOKEN`.
2. When set, render `acme_dns {provider} {token}` into the global options block.
3. When set, install a Caddy build that includes the matching provider module. Caddy's download API makes this a one-liner, e.g.
   `https://caddyserver.com/api/download?p=github.com/caddy-dns/cloudflare`

A useful property of this shape: `acme_dns` in the global options applies to on-demand issuance as well, so the existing `tls { on_demand }` site blocks keep working untouched. The diff stays small.

### Secondary benefit (applies to public deployments too)

DNS-01 is the only challenge type Let's Encrypt will issue **wildcard** certificates for. With it available, a deployment could hold a single `*.code.$HOSTNAME` / `*.dev.$HOSTNAME` pair instead of issuing a certificate per workspace hostname on demand — which meaningfully reduces ACME rate-limit exposure for anyone running many projects.

### Smaller alternative

If a full provider integration is more than you want to own, a documented escape hatch would be enough: have the template `import` an optional user-owned file (e.g. `/etc/caddy/local.caddy`) that survives re-rendering. Users could then add `acme_dns`, supply their own certificates, or set other directives without forking.

### Environment

- Target: single dedicated Ubuntu host, reachable only over a Tailscale tailnet
- DNS: Cloudflare (so `caddy-dns/cloudflare` is the provider in question)
- No inbound ports open, by design

### Happy to contribute

I'm setting this up now and would be glad to send a PR for either approach — just let me know which shape you'd prefer to maintain.

FekyBaz:
Resolves #27\n\n### Summary\n\nAllows hosts without public inbound reachability (Tailscale, WireGuard, private LANs, CGNAT) or deployments requiring DNS-01 ACME challenges (e.g. cme_dns cloudflare <token>) to persist their custom Caddy configuration without maintaining a fork of Caddyfile.tmpl.\n\n### Changes\n\n- infra/templates/Caddyfile.tmpl: \n  - Added import /etc/caddy/conf.d/global/*.caddy in global options block for ACME DNS providers and global TLS directives.\n  - Added import /etc/caddy/conf.d/sites/*.caddy for custom site blocks and reverse proxy endpoints.\n- infra/steps/03-caddy.sh: Automatically ensures /etc/caddy/conf.d/global and /etc/caddy/conf.d/sites exist with proper permissions before template rendering and validation.\n- docs/04-operations/09-deployment-and-operations.md: Documented custom snippet persistence across updates and re-installations.

### DIFF
--- a/docs/04-operations/09-deployment-and-operations.md
+++ b/docs/04-operations/09-deployment-and-operations.md
@@ -116,6 +116,15 @@ flowchart TD
 
 Caddy validates its rendered configuration before replacing the live file. On-demand certificate requests are accepted only for existing project slugs and permitted hostname formats.
 
+### Custom Caddy configuration (`/etc/caddy/conf.d/`)
+
+To support setups such as private LANs, Tailscale tailnets, or DNS-01 ACME challenges (e.g. `acme_dns cloudflare <token>`) without maintaining a fork of `Caddyfile.tmpl`, Remote automatically imports optional custom Caddy snippets:
+
+- `/etc/caddy/conf.d/global/*.caddy`: Imported inside Caddy's global options block. Use this for global ACME DNS plugins, custom certificates, or global TLS options.
+- `/etc/caddy/conf.d/sites/*.caddy`: Imported as additional top-level site blocks. Use this for extra domain names, custom reverse proxies, or dedicated subdomains.
+
+Files in these directories persist across host convergence, updates, and `infra/install.sh` runs.
+
 ## Base-image build
 
 ```mermaid
--- a/infra/steps/03-caddy.sh
+++ b/infra/steps/03-caddy.sh
@@ -8,6 +8,10 @@
 set -euo pipefail
 
 log "Rendering /etc/caddy/Caddyfile for $HOSTNAME"
+# Ensure user snippet directories exist so Caddy can safely import them.
+mkdir -p /etc/caddy/conf.d/global /etc/caddy/conf.d/sites
+chmod 0755 /etc/caddy/conf.d /etc/caddy/conf.d/global /etc/caddy/conf.d/sites 2>/dev/null || true
+
 # Render to a temp file first, validate, then atomically replace. This way a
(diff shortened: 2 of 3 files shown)


FekyBaz:
Closing in favor of #152, which implements the issue's primary proposal (rendered `acme_dns` + provider module build) rather than only the escape hatch. The import-dirs concept here remains available as a follow-up if you want both. No review needed on this one.

PR #148, opened 2026-09-03
fix(auth): restrict first-time credential setup to server loopback

### LINKED ISSUE #65: Move first-time user credential setup to server terminal

FekyBaz:
## Summary

Closes #65

The first-time user credential setup was previously exposed via the public web UI, allowing anyone who accessed the server first to claim administrative privileges before the server owner.

## Changes

1. **Backend Endpoint Security (`/auth/local/claim`)**:
   - Gated the `/auth/local/claim` handler behind loopback/localhost check (`127.0.0.1` / `::1`).
   - Remote requests are rejected with `403 Forbidden` (`admin setup must be performed from the server terminal`).
   - Extended `auth.Status` with `claimAllowed: bool` which is only set to `true` when the server is unclaimed AND the current request is from loopback.

2. **Frontend Protection**:
   - Added `claimAllowed` to `AuthSession` model and API parser.
   - Created [`TerminalSetupRequired.tsx`](file:///frontend/src/ui/auth/TerminalSetupRequired.tsx) to guide remote visitors to complete setup from the terminal (`remote setup`) instead of presenting the claim form.
   - Updated `AuthGate.tsx` to display `TerminalSetupRequired` for remote users if `!auth.claimed && !auth.claimAllowed`.

3. **Testing**:
   - Added `auth_local_handler_test.go` with unit tests verifying:
     - Remote IPs (LAN, public IPv4/IPv6) receive `403 Forbidden`.
     - Loopback callers (`127.0.0.1`, `::1`) are permitted.
     - `/auth/me` properly reflects `claimAllowed: true` only on loopback.

### DIFF
--- a/backend/internal/service/auth/model.go
+++ b/backend/internal/service/auth/model.go
@@ -58,6 +58,11 @@ type Session struct {
 type Status struct {
 	Authenticated        bool           `json:"authenticated"`
 	Claimed              bool           `json:"claimed"`
+	// ClaimAllowed is true only when the local admin credential has not been
+	// set yet and the current request originates from the server's loopback
+	// interface (127.0.0.1 or ::1). Remote visitors receive false so the web
+	// UI does not expose the first-time setup form to arbitrary network clients.
+	ClaimAllowed         bool           `json:"claimAllowed"`
 	LocalAdminConfigured bool           `json:"localAdminConfigured"`
 	GoogleOAuthEnabled   bool           `json:"googleOAuthEnabled"`
 	GoogleClientID       string         `json:"googleClientId,omitempty"`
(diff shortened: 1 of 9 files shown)


kingsoftheweb:
@FekyBaz  I believe this was already handled here? https://github.com/futrx-com/remote.futrx/pull/103

Issue #80, opened 2026-08-28
Pinned chrome version is not compatible with aarch64


Ahmedwaleed22:
Installation is returning the following error on AArch64 machines

<img width="1294" height="410" alt="Image" src="https://github.com/user-attachments/assets/c5e042c7-c3af-4629-865c-5f0e0e6671e2" />

omarabdelfattah:
This still reproduces on aarch64 at the commit that closed it.

d44a0627 fixed the per-arch **path** probing, but the **version comparison**
next to it is still exact:

```sh
if "$browser_bin" --version 2>/dev/null | grep -Fq "$PW_CFT_VERSION"; then
```

Chrome for Testing publishes no `linux-arm64` build (`148.0.7778.96` exists for
`linux64`, `mac-arm64`, `mac-x64`, `win32`, `win64` only), so on aarch64
Playwright installs its own Chromium, which reports `Chromium 148.0.7778.0`.
That never matches the pinned `148.0.7778.96`, so the bake still dies with the
exact message in this issue's title:

```
[3/6] Installing the agent browser and Chromium failed after 1m49s
pinned Chrome for Testing 148.0.7778.96 was not installed
```

The same exact-match exists in `gui-up.sh` and in `stackCheck()`, so even once
the bake passes the launcher and the runtime health probe still reject a
working arm64 browser.

Opened https://github.com/futrx-com/remote.futrx/pull/170 with a fix (prefix match on `major.minor.build`), verified on an
Oracle Ampere A1 — the base image now bakes and publishes. Could this be
reopened?

PR #170, opened 2026-09-07
fix(browser): match Chromium version by prefix so aarch64 installs pass

omarabdelfattah:
## Problem

The base image bake fails 100% of the time on aarch64 at `[3/6] Installing the
agent browser and Chromium`. The browser installs correctly — the version
assertion rejects it.

Chrome for Testing publishes no `linux-arm64` build. From
`known-good-versions-with-downloads.json` for the pinned `148.0.7778.96`:

```
chrome platforms: ['linux64', 'mac-arm64', 'mac-x64', 'win32', 'win64']
```

So on aarch64 Playwright installs **its own** Chromium build, which reports a
different patch component:

```
$ /root/.cache/ms-playwright/chromium-1223/chrome-linux/chrome --version
Chromium 148.0.7778.0
```

`148.0.7778.0` never matches `grep -Fq "148.0.7778.96"`, so the loop never
selects a browser:

```
[3/6] Installing the agent browser and Chromium failed after 1m49s
build failed: agent browser install script: exit status 1
pinned Chrome for Testing 148.0.7778.96 was not installed
```

## Relationship to #80

#80 ("Pinned chrome version is not compatible with aarch64") was closed by
d44a0627, which fixed the per-arch **path** probing in three files but left the
**version comparison** exact in all three. The symptom in #80's title therefore
still reproduces on every aarch64 host.

## Why all three sites

Fixing only the smoke test gets the bake through but leaves the browser
unusable at runtime:

| File | Symptom on aarch64 |
| --- | --- |
| `assets/agent-browser-smoke-test.sh:11` | bake fails at `[3/6]` |
| `assets/gui-up.sh:38` | launcher exits `pinned Chrome for Testing ... is not installed` |
| `install.go:64` (`stackCheck()`) | `provisioner.go:39` never sees a healthy stack, so the on-demand repair path reinstalls the browser on every check |

## The fix

Match the `major.minor.build` prefix, allowing any patch component. This still
rejects a browser from a different milestone — the regression the existing
comments warn about.

`install.go` builds its check in Go, so the prefix is computed and its dots
escaped there.

**`[[:space:]]*` is required, not cosmetic** — `chrome --version` emits a
trailing space:

```
$ chrome --version | cat -A
Chromium 148.0.7778.0 $
```

A fix anchored with a bare `$` silently fails on this.

| `chrome --version` | Result |
| --- | --- |
| `Chromium 148.0.7778.0 ` (arm64 Playwright build, trailing space) | accept |
| `Chromium 148.0.7778.96` (x86_64 Chrome for Testing) | accept |
| `Chromium 147.0.7000.10 ` (wrong milestone) | reject |
| `Chromium 148.0.77780.5 ` (prefix bleed) | reject |

## Verification

Oracle Ampere A1, Ubuntu 24.04 arm64 builder container.

Before:

```
[3/6] Installing the agent browser and Chromium failed after 1m49s
pinned Chrome for Testing 148.0.7778.96 was not installed
```

After:

```
[3/6] Installing the agent browser and Chromium finished in 1m42s
[4/6] Installing the browser IDE finished in 33s
[5/6] Finalizing the builder container finished in 1s
[6/6] Publishing the reusable workspace image finished in 3m38s
done. published "futrx-remote-dev-base".
```

Runtime probe, in a container launched from the resulting image:

```
version: [Chromium 148.0.7778.0 ]
old stackCheck (grep -Fq "148.0.7778.96") -> exit 1
new stackCheck (prefix regex)             -> exit 0
```

`gofmt` clean, `go vet` clean, and
`go test ./internal/integration/containers/browser/` passes.
`browser_test.go` derives its expectation from `browserStackCheck()`, so no
test change was needed. `bash -n` passes on both shell assets.

Note: `go build ./...` fails on this tree with
`static.go:8:12: pattern public: no matching files found`, but it fails
identically on unpatched `main` — it needs the frontend built first, unrelated
to this change.

## Possibly worth a follow-up

- `--version` is called with `2>/dev/null`, so a browser that is present but
  fails to exec (missing shared library) produces empty stdout and is reported
  as "was not installed" — indistinguishable from a genuinely missing install.
  That sends you debugging the download step instead of the assertion.
- `vendors/publish-playwright-assets.sh` publishes only `chrome-linux64.zip`,
  `chrome-headless-shell-linux64.zip` and `ffmpeg-linux.zip`. Release
  `vendors-playwright-1.60.0` has no arm64 assets, so on aarch64 the vendored
  fallback can never succeed — if Google's CDN geo-blocks the host (the exact
  case that fallback exists for), an arm64 install has no recovery path.
  `chromium-linux-arm64.zip` and `ffmpeg-linux-arm64.zip` do exist on the
  Playwright CDN.


kingsoftheweb:
@omarabdelfattah  thanks for including the detailed "Relationship" explanation

PR #176, opened 2026-09-07
fix(infra): validate bootstrap ref before root gate in install.sh

FekyBaz:
## What

Fixes the remaining red `infra shell tests` leg (`FAIL: curl-mode core install.sh gave an
unclear immutable-ref error` in `qa-scripts-test.sh`).

## Root cause

Same pattern as #171: in the curl-piped bootstrap block of `infra/install.sh`, the root gate
ran before the `--ref` validation, so non-root CI saw "needs root" instead of `--ref must be
a full 40-character commit SHA`. (Direct-file invocation was unaffected — only the
`bash -s -- ... < install.sh` path.)

## Changes

- Pure reorder (plus comment): parse bootstrap args, validate the SHA, then demand root.
  No behavior change for real deployments.

## Testing

- Reproduced the exact CI failure locally via Git Bash before the fix; after it the
  curl-mode contract holds.
- Full `bash infra/tests/qa-scripts-test.sh` passes locally end to end (exit 0).

### DIFF
--- a/infra/install.sh
+++ b/infra/install.sh
@@ -47,6 +47,23 @@ set -euo pipefail
 # files on disk.
 INFRA_DIR_PROBE="$( cd "$( dirname "${BASH_SOURCE[0]:-}" 2>/dev/null || echo . )" >/dev/null 2>&1 && pwd || true )"
 if [ -z "$INFRA_DIR_PROBE" ] || [ ! -d "${INFRA_DIR_PROBE}/steps" ]; then
+    # Parse bootstrap-only values before touching either checkout. The full
+    # argument loop below parses them again after we re-exec from disk.
+    BOOTSTRAP_TOKEN="${GITHUB_TOKEN:-}"
+    BOOTSTRAP_REF=""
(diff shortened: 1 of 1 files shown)


kingsoftheweb:
@FekyBaz   Btw,  since you're getting very accustomed, and familiar with the shell scripts, do you think we could refactor it to become a little more organized modular way? like split the files in a more organized way, while consolidating repeated scripts in blocks/functions?

FekyBaz:
Good call — done in #182. I pulled the blocks duplicated across install.sh, update.sh, deploy-app.sh, and upgrade-workspaces.sh (log/warn/ok/err output, the root guard, and the full-SHA validation) into a single sourced infra/lib/common.sh, plus small die and script_dir helpers. The curl-piped bootstrap block stays self-contained since no checkout exists there yet. qa-scripts-test.sh passes and the rest of the shell suite matches the clean tree. Happy to go further (per-area lib files) if you want it in smaller steps.

PR #182, opened 2026-09-08
refactor(infra): consolidate shared shell helpers into lib/common.sh

FekyBaz:
Responds to the refactor request on #176. Moves the helpers duplicated across the infra entry points into one sourced file, infra/lib/common.sh: log/warn/ok/err output, require_root guard, validate_full_sha check, plus new die and script_dir helpers. install.sh, update.sh, deploy-app.sh, and upgrade-workspaces.sh now source it instead of redefining the same blocks. The curl-piped bootstrap block in install.sh stays self-contained (no checkout on disk there yet) with a comment saying so. Behavior is unchanged except upgrade-workspaces root/usage errors now use the shared plain message instead of the styled err prefix. deploy-app.sh keeps its FUTRX_INSTALL_DIR test-harness root bypass and check ordering. Verification: bash -n clean on all touched files; infra/tests/qa-scripts-test.sh passes (covers the pinned --ref error in direct and curl-piped modes); full infra shell suite results are byte-identical to the clean tree (install-migration symlink and go-toolchain versions.env failures pre-exist on this machine). Happy to split further (e.g. per-area lib files) if you want this in smaller steps.

### DIFF
--- a/infra/deploy-app.sh
+++ b/infra/deploy-app.sh
@@ -26,6 +26,10 @@
 # Test-only overrides: FUTRX_INSTALL_DIR, FUTRX_SERVICE_NAME, FUTRX_SERVICE_PORT.
 set -euo pipefail
 
+SCRIPT_INFRA_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+# shellcheck source=lib/common.sh
+. "$SCRIPT_INFRA_DIR/lib/common.sh"
+
 usage() {
     sed -n '2,/^set -euo pipefail$/ { /^set -euo pipefail$/d; s/^# \{0,1\}//p; }' "$0"
 }
@@ -61,9 +65,8 @@ fi
 # demanding root there would make the script untestable in CI. The
 # missing-installation check above intentionally runs first so a bad path
 # reports the actionable error instead of a misleading root demand.
-if [ -z "${FUTRX_INSTALL_DIR:-}" ] && [ "$EUID" -ne 0 ]; then
-    echo "this application deployer needs root; rerun with sudo" >&2
-    exit 1
+if [ -z "${FUTRX_INSTALL_DIR:-}" ]; then
+    require_root "this application deployer"
 fi
 if ! systemctl cat "$SERVICE_NAME" >/dev/null 2>&1; then
     echo "$SERVICE_NAME is not installed; run infra/install.sh first" >&2
@@ -81,7 +84,6 @@ for command_name in git npm go systemctl; do
     }
 done
 
-SCRIPT_INFRA_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 # shellcheck source=lib/release-version.sh
 . "$SCRIPT_INFRA_DIR/lib/release-version.sh"
 # shellcheck source=lib/update-progress.sh
--- a/infra/install.sh
+++ b/infra/install.sh
@@ -60,6 +60,7 @@ if [ -z "$INFRA_DIR_PROBE" ] || [ ! -d "${INFRA_DIR_PROBE}/steps" ]; then
     # Validate the ref before demanding root so a malformed --ref reports the
     # actionable error instead of a misleading root demand (and so the
     # curl-piped contract in qa-scripts-test.sh holds for non-root runners).
+    # Inline (not via lib/common.sh): no checkout exists on disk yet here.
     if [ -n "$BOOTSTRAP_REF" ] && ! printf '%s' "$BOOTSTRAP_REF" | grep -qE '^[0-9a-fA-F]{40}$'; then
         echo "--ref must be a full 40-character commit SHA" >&2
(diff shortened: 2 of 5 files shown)


kingsoftheweb:
@FekyBaz   This is great. But i was wondering if we can take it to the next level when it comes to "modularity"

i was thinking of something like this

infra/
├── config/
│   └── defaults.sh
├── versions.env
├── lib/
│   ├── health-check.sh
│   ├── lxd-host.sh
│   ├── release-version.sh
│   └── systemd.sh
├── steps/
│   ├── host-deps.sh
│   ├── app.sh
│   └── caddy.sh
├── install.sh
├── update.sh
└── deploy-app.sh


This will highly prevent duplication of "defaults/constants" like:

```
readonly FUTRX_DEFAULT_INSTALL_DIR="/opt/remote.futrx"
readonly FUTRX_DEFAULT_SERVICE_NAME="remote.futrx.service"
readonly FUTRX_DEFAULT_SERVICE_PORT="7682"
readonly FUTRX_DEFAULT_BASE_IMAGE="futrx-remote-dev-base"
readonly FUTRX_REPOSITORY_URL="https://github.com/futrx-com/remote.futrx.git"

```
 
 Also for install.sh
 
  We started by trying to require the steps in a modular way but it has a weakness now.sourcing those files executes them immediately and relies heavily on shared global variables. The control flow is therefore less explicit.
  
   So I was also thinking of something like this:
   
   ```
   main() {
   
    remote_parse_install_arguments "$@"
    remote_validate_host
    remote_checkout_repository
    remote_load_configuration

    remote_install_host_dependencies
    remote_install_application
    remote_configure_caddy
    remote_install_backend_service
    remote_build_base_image
    remote_harden_ssh
    remote_install_lxc_ipv4_healing

    remote_verify_installation
    remote_print_install_summary
}

main "$@"
```

and for the other code blocks, like "self bootstrap" , it could be like this

```

remote_self_bootstrap() {
    local infra_probe

    infra_probe="$(
        cd "$(dirname "${BASH_SOURCE[0]:-}" 2>/dev/null || echo .)" \
            >/dev/null 2>&1 &&
        pwd || true
    )"

    # Running from an existing checkout: continue normally.
    if [ -n "$infra_probe" ] && [ -d "$infra_probe/steps" ]; then
        return
    fi

    # Running through curl | bash:
    # 1. Validate bootstrap arguments.
    # 2. Ensure Git is installed.
    # 3. Clone or update the repository.
    # 4. Select the requested commit.
    # 5. Replace this process with the checked-out installer.
    remote_bootstrap_checkout "$@"

    exec bash "$REMOTE_INSTALL_DIR/infra/install.sh" "$@"
}


main() {
    remote_self_bootstrap "$@"

    # These now exist because bootstrap has either returned from an existing
    # checkout or re-executed this file from the newly created checkout.
    remote_load_install_modules
    remote_parse_install_arguments "$@"
    remote_load_configuration
    remote_validate_host

    remote_select_checkout
    remote_install_host_dependencies
    remote_install_application
    remote_configure_caddy
    remote_install_backend_service
    remote_build_base_image
    remote_harden_ssh
    remote_install_lxc_ipv4_healing

    remote_verify_installation
    remote_print_install_summary
}

main "$@"
```

FekyBaz:
@kingsoftheweb Good direction overall — explicit `main()` orchestration, function-only library files, and a self-bootstrap path are exactly how this should end up. A few notes so the work lands cleanly:

**Much of the layout already exists.** `infra/` already has per-area `lib/` files (health-check, lxd-host, release-version, …), numbered `steps/` (00-checkout through 07-lxc-ipv4-heal), and `versions.env`. Before reorganizing, please take a fresh look at `qa` — your proposal's `lib/health-check.sh`, `lxd-host.sh`, `release-version.sh` and the steps list map 1:1 onto what's there. The real gaps vs your sketch are: (a) no `config/defaults.sh` centralizing the `FUTRX_DEFAULT_*` constants (worth doing — they're currently scattered), and (b) no explicit `main()` entry points.

**Keep it staged — don't balloon this PR.** #182 is deliberately a behavior-preserving foundation (shared helpers only, verified byte-identical shell-suite output). I'd suggest: merge this first, then do the orchestration as a follow-up: function-only convention for `steps/` + `main()` in install/update/deploy + the `config/defaults.sh` extraction. Smaller reviews merge faster.

**Two constraints for the follow-up:**
1. The `curl | bash` path must keep working — that's why #182 deliberately left the bootstrap block in `install.sh` self-contained. Your `remote_self_bootstrap` handles exactly this, good — just make sure the `steps/`-presence probe matches the real layout (`steps/`, not `install_modules` naming drift).
2. For the sourcing-executes-immediately weakness you flagged: the cheap fix is the standard guard (`[[ "${BASH_SOURCE[0]}" == "${0}" ]] || return` at top of library files, `main "$@"` at the bottom of entry points). No framework needed.

Happy to implement the follow-up (defaults centralization + `main()` entry points) once #182 lands, if you want to split the work that way.

kingsoftheweb:
@FekyBaz  Sure, let's do it like that. I will merge this one

PR #184, opened 2026-09-09
feat(chat): simplify approval request details

### LINKED ISSUE #156: Codex agent message is not simple for all users
We need better message formatting for the Codex agent questions

Image

OsamaAlSabry:
## Summary

- Simplify Codex approval request details by showing the command, action, and reason as a readable summary.
- - Keep the complete request payload available behind an expandable `Show full request details` section.
- - Preserve the existing approval and cancellation actions without changing their response semantics.
- - Add focused unit coverage for summary field ordering, fallback keys, structured values, and missing fields.
Closes #156 

## Scope note

The maintainer requested a `Deny, do this instead` action that would send alternative instructions back to the agent. The current `ChatInteractionIntent` and `ChatInteractionService` do not expose a response contract for that flow, so this PR intentionally does not add a non-functional button or change the backend protocol. That follow-up requires an agreed wire-level response shape.

## Testing

- `npm test -- --test-name-pattern='approval summary'`
- - `npm run build`
- - `git diff --check`
The frontend test suite passed after installing the repository dependencies: 49 tests passed, 0 failed. The production build completed successfully.

## UX and accessibility

The summary uses a responsive grid that collapses naturally on narrow screens, wraps long values, and keeps the full payload available for inspection. Existing button labels and actions remain unchanged.


frontend/src/ui/chat/interactions/approvalSummary.ts
```diff
@@ -0,0 +1,37 @@
+export interface ApprovalSummaryField {
+  label: string;
+  value: string;
+}
+
+const FIELD_SOURCES: Array<{ label: string; keys: string[] }> = [
+  { label: "Command", keys: ["command", "cmd"] },
```
kingsoftheweb, line 7: @OsamaAlSabry 

 There is a list of command actions, and each one of them has a more user-friendly values than the pure bash-type of commands.  I believe you meant to use this one in the screenshot, correct?

Image

PR #189, opened 2026-09-10
refactor(infra): centralize defaults + explicit main() entry points

FekyBaz:
Follow-up to #182 as discussed there.Combine two changes:1. infra/config/defaults.sh centralizes the shared install dir, service name/port, base image and repo URL constants (env overrides preserved exactly, including deploy-app's PORT vs SERVICE_PORT wiring).2. All 8 steps/*.sh are now function-only libraries (step_NN_*) called explicitly from main(); the 4 entry points gained main() plus source-safe guards. The guard defaults to executing because BASH_SOURCE is unset under set -u when bash reads from stdin - the existing qa-scripts-test curl-mode case caught the naive version during development.Two behavior notes: step-file sourcing no longer executes (remote-cli-test updated with the one-line call, same subshell-abort semantics); QA operator scripts untouched.Verified here: bash -n clean on all 14 files, qa-scripts-test (full) + remote-cli-test green, host-agent-install-test static contracts hold. The Go-dependent part of host-agent-install-test needs CI (no Go toolchain on this machine).

### DIFF
--- a/infra/install.sh
+++ b/infra/install.sh
@@ -40,6 +40,7 @@
 
 set -euo pipefail
 
+main() {
 # ───────────────── self-bootstrap (curl|bash mode) ─────────────────
 # When piped from curl, BASH_SOURCE points at /dev/stdin and there are no
 # sibling steps/ or templates/. Install git, clone the repo to the canonical
@@ -158,6 +159,8 @@ fi
 INFRA_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 # shellcheck source=lib/common.sh
 . "$INFRA_DIR/lib/common.sh"
+# shellcheck source=lib/../config/defaults.sh
+. "$INFRA_DIR/config/defaults.sh"
 HOSTNAME=""
 SKIP_DNS_CHECK=0
 GOOGLE_CLIENT_ID=""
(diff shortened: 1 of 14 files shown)


infra/install.sh
```diff
@@ -40,6 +40,7 @@
 
 set -euo pipefail
 
+main() {
```
kingsoftheweb: @FekyBaz  this is not very modular. it is pretty much the same as it was, but wrapped in main().

We need to follow this structure , or something similar

https://github.com/futrx-com/remote.futrx/pull/182#issuecomment-5602988850

kingsoftheweb (changes requested):

FekyBaz:
@kingsoftheweb Done — all four entry points are now split into named phase functions behind an explicit `main()`, per your modularity request:

- `install.sh`: `remote_self_bootstrap / parse_install_arguments / load_configuration / migrate_legacy_install / select_checkout / validate_host / converge_host / print_install_summary`
- `update.sh`: `load_configuration / parse_update_arguments / migrate_legacy_install / refresh_checkout / detect_hostname / converge_update`
- `deploy-app.sh`: `load_configuration / parse_deploy_arguments / validate_deploy / build_application / verify_deployment`
- `upgrade-workspaces.sh`: `load_configuration / parse_workspace_arguments / validate_host / rebake_base_image / migrate_containers`

No behavior change: same statements in the same order (verified via diff), `curl | bash` stdin mode and the sourced-by-tests guard untouched, shared helpers already live in `infra/steps/*.sh` + `infra/config/defaults.sh`.

Verification: `bash -n` clean on all infra files, `qa-scripts-test.sh` and `remote-cli-test.sh` both green.

FekyBaz:
@kingsoftheweb Your changes-requested review (14:41 UTC) landed about an hour before the phase-split push (f6b64af, ~15:48 UTC) — all four entry points are now named phase functions behind explicit main(), same statements, same order, tests green. Requesting a re-review whenever you have a moment.

kingsoftheweb (approved):

PR #199, opened 2026-09-11
feat(applications): installable image catalog, installer and HTTP API

Ahmedwaleed22:
Adds the foundation of the installable-application subsystem: a catalog of images that a server administrator can install into a container, and the service, persistence and HTTP layers around it.

An image is a directory under images/ holding an image.json and an install script. The catalog is compiled into the server binary by a small module at the repository root, because go:embed reaches only downwards and the catalog is deliberately kept where someone adding an app finds it.

Two kinds of image land here:

  service  runs software on a port under systemd, reachable on the host
           through an LXD proxy device. Global scope gets a dedicated
           container; project scope installs into the project's own.
  tool     provisions software into the project's container and exposes
           nothing — a CLI, a mount, an agent. Project scope only.

Layering follows the existing ports-and-adapters split:

  service/applications          policy: install, lifecycle, scope, env
  integration/containers/...    the lxc side: install script, systemd, proxy
  integration/hosttools         host executables an image declares, checksum-pinned
  stores/fileapplications       installed-instance persistence
  transport/http/handlers       routes and authorization

Global routes are admin-only; per-project routes are delegated by ProjectHandler, which has already enforced membership.

No frontend in this PR — the Applications UI follows on top of it. The handbook under docs/dev/installable-images describes the whole subsystem, including the halves that land in the PRs stacked on this one.

Split out of #50.


applications/README.md
```diff
@@ -0,0 +1,139 @@
+# Application images
+
```
kingsoftheweb, line 2: rename to "applications"

and everywhere it needs to be referred to as "application(s)"

docs/dev/installable-images/01-overview.md
```diff
@@ -0,0 +1,300 @@
+# 01 — Overview
+
+## What the catalog is
+
+The catalog is a directory of subdirectories. Each subdirectory is one
+installable image:
+
+```
+images/
```
kingsoftheweb: rename to "applications"

docs/dev/installable-images/01-overview.md
```diff
@@ -0,0 +1,300 @@
...
+  docs/              ← this documentation (reserved name, not an image)
+  mysql/
+    image.json       metadata
+    install.sh       provisioner, run inside a container
+    ui/              browser extension (optional)
+    plugin/          Go backend, compiled and run on the host (optional)
```
kingsoftheweb: as discussed, will be renamed to "backend"

docs/dev/installable-images/01-overview.md
```diff
@@ -0,0 +1,300 @@
...
+## What the catalog is
+
+The catalog is a directory of subdirectories. Each subdirectory is one
+installable image:
+
+```
+images/
+  docs/              ← this documentation (reserved name, not an image)
+  mysql/
+    image.json       metadata
```
kingsoftheweb: this will be "application.json" -> entry point

backend/internal/integration/containers/applications/host_tools_test.go
```diff
@@ -0,0 +1,114 @@
+package applications
```
kingsoftheweb, line 1: do we still need this?

backend/internal/integration/containers/applications/catalog_fixture_test.go
```diff
@@ -0,0 +1,64 @@
+package applications
```
kingsoftheweb, line 1: check if this is still required?

backend/internal/integration/containers/applications/infra_payload.go
```diff
@@ -0,0 +1,125 @@
+package applications
```
kingsoftheweb, line 1: do we actually need this?

If this is actually pertaining to the global container that is generated on `ScopeGlobal` then :

1- it needs to be renamed to properly match the intent.
2- this could also have a UI for updating its resources limits (the same as how we do it per project at the moment). But since this would be a "global" container, then it could live under The Global settings screen, and it should have the same behaviour as the "per-project" resources settings

backend/internal/integration/containers/applications/registry.go
```diff
@@ -0,0 +1,303 @@
...
+	"strings"
+	"sync"
+
+	"futrx.local/catalog"
+
+	"github.com/futrx-com/remote.futrx.com/internal/integration/hosttools"
+	svc "github.com/futrx-com/remote.futrx.com/internal/service/applications"
+)
+
+// Registry is an in-memory, validated view of an image catalog.
```
kingsoftheweb: Please confirm that we read/write apps and their meta data in disk. and if there is any in-memory action, then please explain it

backend/internal/integration/hosttools/hosttools.go
```diff
@@ -0,0 +1,322 @@
+// Package hosttools installs the host-side executables an installable image
```
kingsoftheweb: review if we actually need this. and if it is actually a "utility" set of methods. then move them to the "shared" layer https://github.com/futrx-com/remote.futrx/tree/main/backend/internal/shared

docs/dev/installable-images/01-overview.md
Ahmedwaleed22: Done

Ahmedwaleed22:
@kingsoftheweb  Here is the new application structure

```
applications/<your-app-id>/
  application.json        ← the manifest (required)
  install.sh               ← installer script (service/tool only)
  infra/                   ← infrastructure: container-side source (optional)
  infra.tar.gz              ← packaged infra/, built by your own package.sh (optional)
  backend/                 ← backend: Go source compiled & run by the server (optional)
    main.go
  ui/                      ← UI: browser extension (optional)
    scripts/main.js
    style/*.css
    views/*.html
    assets/*               ← your assets (icons, images, etc.)
```

backend/internal/integration/containers/applications/host_tools_test.go
Ahmedwaleed22: host_tools_test.go — yes, still needed. It tests HostTool/ensureHostTools, which is live code: an app can declare a host-side binary (checksum-pinned download), and installer.go installs/repairs it before touching the guest. Not dead weight.

backend/internal/integration/containers/applications/catalog_fixture_test.go
Ahmedwaleed22: yes, still needed. It's not a standalone test, it's the shared fixture (fixtureCatalog(), testRegistry()) that six other test files in the package depend on (registry_test.go, registry_ui_test.go, registry_backend_test.go, installer_test.go, packages_test.go, and host_tools_test.go itself). Deleting it breaks the whole test suite.

backend/internal/integration/containers/applications/infra_payload.go
Ahmedwaleed22: infra_payload.go is not the global container — it's unrelated to ScopeGlobal. What it actually does: some applications (like s3disk) need to ship container-side source as a nested Go module, which go:embed can't traverse. So the app author packages that source into infra.tar.gz, and this file validates/extracts that archive into the target container right before install.sh runs. It's a build-and-package mechanism, not a runtime container. So the premise "rename it to match the global-container intent" doesn't apply — it already matches its actual intent (shipping an app's infra/ source). It's still needed for any app with a nested-module container payload.

backend/internal/integration/containers/applications/registry.go
Ahmedwaleed22: Confirmed. Reads only happen at load time: the built-in catalog comes from the embedded binary (//go:embed applications), uploaded packages come from real disk under $DATA_DIR/app-packages/. Writes only happen for uploads — PackageStore.install() stages the archive, validates it through the same loadApplication path the embedded catalog uses, then atomically renames it into place (plus a meta/<id>.json sidecar with who/when/from-what-archive).

The in-memory part is Registry.view: a full snapshot built by Reload(), which runs once at startup and once after every upload/remove. Every read after that (List, Get, Script, UIAsset, BackendSource) hits only that cached struct under r.mu — no disk I/O per request.

Side note: installed instance state (which app is installed where, running/stopped) isn't in this file at all — that's stores/fileapplications, writing JSON under $DATA_DIR/applications/. Let me know if that's actually what this comment was after.

backend/internal/integration/hosttools/hosttools.go
Ahmedwaleed22: Still needed — it's live, not dead code: installer.go calls Ensure before running the guest install script, and registry.go calls Validate at catalog load.

On moving it to shared/: it's not a pure utility as written, so I'd hold off. Both current shared/ packages (output, workspacepath) have zero domain imports; hosttools's public API (Ensure, Validate) is typed directly against svc.HostTool/svc.HostToolDownload from service/applications, and it has exactly one consumer today (containers/applications). Dropping it into shared/ as-is would mean shared/ importing a specific domain package.

It could get there with a moderate refactor — take primitives (name, version, URL, sha256, compression) instead of the svc.HostTool struct, and have applications translate at the call site. Can spin that out as a follow-up if it's worth doing, rather than folding it into this PR.

docs/dev/installable-applications/01-overview.md
```diff
@@ -0,0 +1,300 @@
...
+
+The catalog is a directory of subdirectories. Each subdirectory is one
+installable application:
+
+```
+applications/
+  docs/              ← this documentation (reserved name, not an application)
+  mysql/
+    application.json       metadata
+    install.sh       provisioner, run inside a container
```
kingsoftheweb: @Ahmedwaleed22  this is part of infra
it should live under infra/
top level file(s) should only be the metadata ones like application.json or a readme.md

The application structure needs to mimic as much as it cans core itself:

infra/
backend/
frontend(ui)/

backend/internal/integration/containers/applications/host_tools_test.go
kingsoftheweb: @Ahmedwaleed22  then it should not live here, it should live with hosttools folder

backend/internal/integration/hosttools/hosttools.go
kingsoftheweb: @Ahmedwaleed22  ok if this is tied to applications, then it shuold not be its own integration layer.

this should be a sublayer under integration/containers/applications. specially that it has one consumer which is continaers/applications.

PR #209, opened 2026-09-13
Feat/project chat lifecycle publishers

kingsoftheweb:


backend/internal/lifecycle/chat_publisher.go
```diff
@@ -0,0 +1,62 @@
...
+
+import "context"
+
+// ChatState identifies a persisted chat lifecycle transition.
+type ChatState string
+
+const (
+	ChatCreated ChatState = "created"
+	ChatUpdated ChatState = "updated"
+	ChatDeleted ChatState = "deleted"
```
Ahmedwaleed22, line 11: @kingsoftheweb 
We might need an event of opening called "read"
since we might need to do an action on opening the chat

like "mark as read" or stop sending notifications about any updates

Ahmedwaleed22:
@kingsoftheweb 
We also need a message publisher we need to have the sender, message content, isRead, and the timestamp

PR #223, opened 2026-09-18
feat(permissions): service-owned permission abstraction

AbdallahMohamedDotnet:
## What

Adds a reusable, service-owned permission layer:

- a code-owned permission **registry** (only developers add permissions);
- an **evaluator** with direct allow/deny, custom roles, deny precedence, exact scope matching, and code-owned compatibility baselines;
- **delegation** with anti-escalation rules, evaluated under the store lock so a concurrent revocation cannot pass a stale check;
- a file-backed **policy store** (`permissions.json`) with an append-only **audit log** (`permission-audit.jsonl`); a failed audit rejects the mutation;
- **actor propagation** through the request context, attached by the auth middleware only after the session and registration checks pass;
- service-side enforcement of two project permissions: `projects.lifecycle.manage` (Start, Stop, Restart, RepairNetwork) and `projects.access.manage` (ListAccess, AddAccess, RemoveAccess).

Developer guide: [`docs/dev/permissions.md`](docs/dev/permissions.md).

## Why

Authorization was scattered across handlers and per-service admin/member checks, so calling a service method directly could bypass the policy. The decision now lives in the owning service, and a direct service call gets the same answer as the HTTP route.

## Compatibility

- No new user or admin type; existing `admin`/`member` roles and project membership are unchanged. Administrators are the root policy and cannot be locked out by an assignment.
- With an empty policy store, behavior is unchanged: both permissions use the project-member baseline (covered by route tests through the real middleware).
- No handler files are modified. No data migration: an absent `permissions.json` is an empty policy.
- Startup fails with an actionable error on a corrupt `permissions.json` or a stored reference to an unregistered permission key, rather than silently dropping a deny.

## Things reviewers should know

- **Denials surface as HTTP 500, not 403.** Handlers are frozen in this delivery and none maps `permission.ErrDenied`, so a denial currently returns `500 {"error":"permission denied"}`. Mapping it is a mechanical follow-up.
- **Handler-level admin gates still apply.** Delegated non-administrators cannot reach routes a handler rejects with a hard-coded `isAdmin` check. Roles, assignments, and bindings are exposed as service methods (`Services.Permissions`) only; there is no route or UI yet.
- **Composition order changed.** Auth is now built before the project service, over a user-service view without removal cleanup (the two share one repository), to keep the graph acyclic without late wiring.
- **Stricter than plan rule 4.** Editing or unbind-deleting a *bound* role also requires the actor to be able to delegate every old and new rule at each binding's scope; otherwise a roles manager could widen a role held by others.
- **Trusted internal callers** use an explicit system context, on an allowlist enforced by an architecture test: agent runs, application container readiness, notification fan-out, user-removal cleanup. A member denied `projects.lifecycle.manage` can still cause a start indirectly (agent run, agent browser).
- Five existing tests in `service_start_test.go` gained an allow-all authorizer in setup because a service with no authorizer now fails closed; their assertions are unchanged.

## Verification

- `go test ./...`, `go vet ./...`, `gofmt -l` clean; `go test -race` on the permission, store, project, and middleware packages.
- Concurrency test: 200 iterations of a concurrent revoke and grant, asserting no grant succeeds after the revocation took effect.
- Tests for the safeguards were checked by temporarily disabling each one and confirming a failure.
- The real `cmd/remote` binary starts and listens with an empty policy, and refuses to start on an unknown key or a corrupt file.
- Frontend: `npm test` (359 pass) and `npm run build`; no frontend files changed.
- **Not run:** `infra/qa/deploy-app.sh <pushed-ref>` against a QA installation (needs a QA server); please run it on this branch.

Base is `qa`, matching how recent contributor PRs land, although CONTRIBUTING.md says to branch from `main`. Happy to retarget.


AbdallahMohamedDotnet (commented):
@kingsoftheweb can you review it ?

backend/internal/service/project/service.go
```diff
@@ -326,7 +330,20 @@ func (s *Service) Delete(ctx context.Context, id ID) error {
...
 func (s *Service) Start(ctx context.Context, id ID) (Meta, error) {
+	if err := s.require(ctx, PermissionLifecycleManage, id); err != nil {
+		return Meta{}, err
+	}
+	return s.start(ctx, id)
+}
+
+// start is Start without authorization, for callers inside this package that
+// are already acting under another permission.
+func (s *Service) start(ctx context.Context, id ID) (Meta, error) {
```
Ahmedwaleed22: calling method `start` while we have `Start` method is confusing, make sure to run code refactorer skill or use a better naming

backend/internal/service/permission_composition.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// projectMembership adapts the project access repository to the permission
+// layer's membership port, so the evaluator does not import the project
+// service.
+type projectMembership struct {
+	access serviceproject.AccessRepository
+}
+
+func (m projectMembership) HasAccess(ctx context.Context, projectID string, email string) (bool, error) {
```
Ahmedwaleed22: We need a better way to handle permissions, for example

```
user.HasPermission("project create")
user.HasPermission("project edit")
```

This opens the door to be able to create custom permissions later to use with the `applications`

for roles it must be

```
user.HasRole("admin")
user.HasRole("project1 reviewer")
```

also we need to tie it with the user not the email so if we add the ability to change the email later it will be still working with the user

Ahmedwaleed22:
@AbdallahMohamedDotnet

Great Job! A small addition that might be helpful in many cases

 We need to use a pub/sub architecture so applications can use a hook to listen to specific events

Needed event: "permission_X_granted_to_user_Y"

**_Example 1_**

> Application "X" is used to notify the user on telegram instead of notifications while app "X" is a global app looking at all the apps User A  has access to and sends notifications to it
> 
> User A gets access to resource B then application X starts notifying the user for it
> 
> User A get access removed to resource B the application stop notifying user A about resource B

_**Example 2**_
> Application "X" creates a public link to share with clients but this has a custom permissions showing how many users are able to generate public links
> 
> User A gets permissions to do it so the app starts dealing with this

samem26:
@Ahmedwaleed22 
It was planned and is going to be finished soon.

PR #228, opened 2026-09-20
feat(applications): add first-class container-side Go programs

Ahmedwaleed22:
## Summary

Add first-class support for Go programs that run inside application containers.

- Separate host plugins into `backend/api/` and container programs into `backend/container/`
- Preserve compatibility with the legacy flat `backend/` layout
- Deterministically package container source and build it inside the target LXD container
- Discover and install `cmd/*` binaries into `/usr/local/bin`
- Centralize the container Go toolchain version and generate module metadata when needed
- Track a source-derived build identity so container code changes trigger reprovisioning even without a manifest version bump
- Keep optional `infra/install.sh` support for additional provisioning
- Update Hello Remote to demonstrate container inspection through its host API and UI
- Expand application documentation and test coverage for the new layout

## Validation

- `go test ./...`
- `cd backend && go test ./...`
- `cd frontend && npm test` — 419 tests passed
- `cd frontend && npm run build`


applications/hello-remote/application.json
```diff
@@ -1,9 +1,9 @@
 {
   "id": "hello-remote",
```
kingsoftheweb, line 2: We need to show also the host tools here as an example.  Since hello remote should be pretty much shown everything that could be done in an application.

applications/hello-remote/application.json
```diff
@@ -1,9 +1,9 @@
...
   "name": "Hello Remote",
-  "description": "Example app: a Go plugin on the server, and a panel in the browser that calls it.",
+  "description": "Example app: explore every frontend extension slot and call its Go plugin and LXD container program.",
   "category": "development",
-  "version": "1",
+  "version": "6",
   "icon": "ui/assets/logo.svg",
   "scopes": ["global", "project"],
```
kingsoftheweb, line 8: We also should be adding a service we should be adding everything that is found in the application model.

 https://github.com/Ahmedwaleed22/remote.futrx/blob/2e5bf4dea1e098442fe6f2b9e06579c458f97293/backend/internal/service/applications/model.go#L159-L160

applications/hello-remote/infra/install.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+# OPTIONAL APPLICATION-SPECIFIC INSTALLATION TEMPLATE
```
kingsoftheweb: I need to understand something and I want you to confirm it. This only is tied to the LXD container. So this give you the shell access to the container while host tools asks to install, uh, tools/applications on the host level. Is that it or it's different?

docs/dev/installable-applications/01-overview.md
```diff
@@ -71,9 +71,9 @@ optional and independent:
 
 | Capability | Declared by | What Remote does |
 |---|---|---|
-| Infrastructure | `infra/install.sh`, or an `install` path inside `infra/` | Provisions software in a container |
+| Infrastructure | `infra/install.sh`, an `install` path inside `infra/`, or `backend/container/` | Provisions software in a container |
 | Network exposure | infrastructure plus `port.internal` | Allocates a host port and adds an LXD proxy device |
-| Backend | `backend/` | Compiles and runs the Go backend on the host |
+| Backend | `backend/api/` | Compiles and runs the Go backend on the host |
 | UI | `ui/` | Loads the browser extension |
 | Skills | `skills/*/SKILL.md` | Publishes skills to the target project |
```
kingsoftheweb, line 78: Is this still, uh, applicable? The DU actually provide the ability to provide custom skills? If so, then it should be part also of the hello remote application.

docs/dev/installable-applications/04-install-scripts.md
```diff
@@ -14,6 +21,10 @@ lxc exec <container> --env APP_INTERNAL_PORT=3306 --env … -- bash -s
 
 It receives:
 
+- `APP_APPLICATION_ID`, `APP_APPLICATION_NAME`, and
```
kingsoftheweb, line 28: Is this still applicable and if so then we should need to also show it as an example. Inside the files of the application of hello remote.

applications/hello-remote/ui/scripts/containerPanel.js
```diff
@@ -0,0 +1,58 @@
+export function mountContainerPanel(host, backend, target, isDisposed) {
```
kingsoftheweb, line 3: Confirm please that, uh, the UI is well documented. The API for the UI is well documented. I did see the slots file under the documentation, but I want to make sure that actually has all the information. For how to use the UI API properly.

applications/hello-remote/backend/api/main.go
```diff
@@ -28,36 +23,40 @@ import (
 const visitsFile = "visits.json"
 
 type backend struct {
-	mux *appplugin.Mux
+	mux              *appplugin.Mux
```
kingsoftheweb: The naming of multiplexer as MUX is extremely confusing. This is more of a router and a handler for the endpoints. So we need a better name.

  And as far as I remember, this also will need to be removed and fixed in core. Because I remember seeing MUX somewhere else.

backend/internal/config/constants/application_container.go
```diff
@@ -0,0 +1,5 @@
+package constants
+
+// ApplicationContainerGoVersion is the toolchain Remote installs inside an
+// application container when building backend/container/ programs.
+const ApplicationContainerGoVersion = "1.24.7"
```
kingsoftheweb: why we are pinning down the version here? I don't even think this has any usage anywhere.

backend/internal/integration/containers/applications/container_install.go
```diff
@@ -0,0 +1,111 @@
...
+// containerBuildMarkerDir holds one marker file per application, recording the
+// build its binaries came from.
+//
+// The marker is what makes installing idempotent, and it is deliberately not a
+// question asked of the binary. Doing that would oblige every container program
+// to implement a --version flag and carry a version variable for the build to
+// stamp — an unwritten contract that costs a full toolchain download and rebuild
+// on every install and start when an author does not know to satisfy it. A file
+// the server writes and reads needs nothing from the program at all.
+const containerBuildMarkerDir = "/usr/local/lib/remote"
```
kingsoftheweb: This has no usage anywhere. Could you review and make sure we don't have this hanging "unused" constants? You can run the refactor skill against a specific folder. It will help with that.

  You can ask the agent to run it against a specific folder and make sure that it removes anything that is not used.

applications/hello-remote/infra/install.sh
Ahmedwaleed22: No, there is no way to install something on host

this was removed

docs/dev/installable-applications/01-overview.md
Ahmedwaleed22: Okay will add it

applications/hello-remote/infra/install.sh
Ahmedwaleed22: Host tools install packages on the host,
while install.sh installs inside the container, same as any script inside "infra"

```
"hostTools": [
    {
      "name": "restic",
      "version": "0.18.0",
      "downloads": {
        "amd64": {
          "url": "https://github.com/restic/restic/releases/download/v0.18.0/restic_0.18.0_linux_amd64.bz2",
          "sha256": "98f6dd8bf5b59058d04bfd8dab58e196cc2a680666ccee90275a3b722374438e",
          "compression": "bzip2"
        },
        "arm64": {
          "url": "https://github.com/restic/restic/releases/download/v0.18.0/restic_0.18.0_linux_arm64.bz2",
          "sha256": "ce18179c25dc5f2e33e3c233ba1e580f9de1a4566d2977e8d9600210363ec209",
          "compression": "bzip2"
        }
      },
      "versionArgs": ["version"]
    }
  ],
```

Here is the syntax

applications/hello-remote/application.json
Ahmedwaleed22: Good call — I added a checksum-pinned restic host tool example for both amd64 and arm64, including decompression and version verification.

applications/hello-remote/application.json
Ahmedwaleed22: Agreed. Hello Remote now demonstrates the complete application model: service, port, health check, connection metadata, environment options, base image, host tools, UI, backend, container commands, and skills. I also added a test to keep that coverage from regressing.”

docs/dev/installable-applications/01-overview.md
Ahmedwaleed22: Yes, project applications support custom skills. I added a hello-remote-inspector skill so Hello Remote demonstrates the full publishing flow.

docs/dev/installable-applications/04-install-scripts.md
Ahmedwaleed22: Done. The install script now uses the provided APP_* and HELLO_* variables, writes root-only configuration, provisions the systemd service idempotently, registers it with the idle-workspace probe, and waits for it to become healthy.

applications/hello-remote/ui/scripts/containerPanel.js
Ahmedwaleed22: Confirmed. The full frontend API is documented in 06-extension-api.md, and I checked it against the implementation. I also expanded Hello Remote’s explorer to demonstrate the remaining upload claim() behavior, so the example now covers every exposed API capability.

applications/hello-remote/backend/api/main.go
Ahmedwaleed22: Agreed, Mux was unnecessarily confusing here. I renamed it to Router throughout the public SDK, Hello Remote, the fixture plugin, tests, and documentation.

backend/internal/config/constants/application_container.go
Ahmedwaleed22: This constant is in use, so I kept it. It controls both the Go toolchain downloaded by the container build script and the generated module’s Go version. I updated the comment to make that synchronization and reproducibility purpose clearer.

backend/internal/integration/containers/applications/container_install.go
Ahmedwaleed22: I traced this one as well. `containerBuildMarkerDir` is used to derive each application’s marker path and create its parent directory in the generated installer. It prevents unnecessary toolchain downloads and rebuilds, so it isn’t dead code and should remain.

